PR #34, opened 2026-01-31
feat(core): refine imports and plugin validation

avifenesh:
## Summary
- Update markdown import/XML parsing to skip code spans/blocks and ignore plain @mentions
- Add import graph checks (missing/cycle/depth), plugin validation rules, and repo .agnix.toml config
- Auto-load .agnix.toml and normalize excludes; adjust docs example imports

## Test Plan
- cargo test
- cargo run --bin agnix -- .

### DIFF
--- a/crates/agnix-core/src/rules/plugin.rs
+++ b/crates/agnix-core/src/rules/plugin.rs
@@ -109,7 +109,8 @@ impl Validator for PluginValidator {
         };
 
         if config.is_rule_enabled("CC-PL-003") {
-            if !is_valid_semver(&schema.version) {
+            let version = schema.version.trim();
+            if !version.is_empty() && !is_valid_semver(version) {
                 diagnostics.push(
                     Diagnostic::error(
                         path.to_path_buf(),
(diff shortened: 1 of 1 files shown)


avifenesh:
Thanks for the review! Addressing each point:

1) @import path traversal: Per knowledge-base/standards/claude-code-HARD-RULES.md (3.2), @import allows relative and absolute paths (and ~). This implementation follows that spec; restricting to project scope would be a breaking deviation. If you want an optional restriction flag in config, I can add it in a follow-up.
2) Home directory resolution: Only ~\/ or ~\\ is expanded; if home is unavailable the path remains unresolved and triggers CC-MEM-001, which is consistent with the 'path must exist' rule.
3) Config discovery: Current behavior is load .agnix.toml from target/cwd or fall back to defaults. There is no spec requirement to hard-fail or walk ancestors, so I kept behavior conservative; happy to add ancestor search if desired.
4) Import file read errors: validate_project already emits file read errors for directly validated files. For imported files, we currently treat unreadable as missing (CC-MEM-001). If you'd like a distinct diagnostic, I can add.
5) SemVer: CC-PL-003 in knowledge-base/VALIDATION-RULES.md specifies ^\d+\.\d+\.\d+$, so regex is intentional.

avifenesh:
Fixed: CC-PL-003 now skips empty version strings (so only CC-PL-004 fires). Added test coverage for empty version.

PR #257, opened 2026-02-05
fix(vscode): prevent client lifecycle races

### LINKED ISSUE #241: VS Code: race condition on global client variable during rapid config changes
## Description

The VS Code extension uses a module-level mutable variable for the LSP client:

\`\`\`typescript
let client: LanguageClient | undefined;  // line 16
\`\`\`

Multiple async functions mutate this without synchronization:
- \`startClient()\` (line 521) — sets \`client = new LanguageClient(...)\` at line 591
- \`stopClient()\` (line 617) — sets \`client = undefined\` at line 620
- \`restartClient()\` (line 625) — calls stop then start

The \`onDidChangeConfiguration\` handler (lines 490-512) can trigger \`stopClient()\` or \`restartClient()\` on any config change. If two configuration changes fire in rapid succession (e.g., user edits settings.json), concurrent async operations can:

1. Start two client instances (first \`startClient\` hasn't finished when second begins)
2. Stop a client that's already being stopped
3. Leak the old client reference when \`client\` is overwritten

## Suggested fix

Add a serialization guard:

\`\`\`typescript
let clientMutex: Promise<void> = Promise.resolve();

async function restartClient(): Promise<void> {
    clientMutex = clientMutex.then(async () => {
        await stopClient();
        await startClient();
    });
    return clientMutex;
}
\`\`\`

## Files

- \`editors/vscode/src/extension.ts\` lines 16, 490-512, 521-632

avifenesh:
## Summary
- serialize VS Code LSP client lifecycle transitions behind a dedicated `ClientLifecycleController`
- guard stale start operations with generation-based intent checks and active-client detection
- add unit coverage for overlapping lifecycle operations and stale startup scenarios
- update VS Code extension changelog with issue #241

## Test Plan
- `cargo check`
- `cd editors/vscode && npm run pretest && npm run test:unit`
- `npm test` currently fails on `main` and this branch due a pre-existing script path bug (`out/test/runTest.js`)

## Related Issues
Closes #241

### DIFF
--- a/editors/vscode/src/extension.ts
+++ b/editors/vscode/src/extension.ts
@@ -371,12 +371,7 @@ export async function activate(
     },
     isClientActive: (runningClient: LanguageClient) => runningClient.isRunning(),
     startClient: () => startClientInternal(),
-    stopClient: async (runningClient: LanguageClient) => {
-      await runningClient.stop();
-      if (client === runningClient) {
-        client = undefined;
-      }
-    },
+    stopClient: (runningClient: LanguageClient) => runningClient.stop(),
   });
 
   context.subscriptions.push(
@@ -1449,11 +1444,5 @@ function updateStatusBar(
 }
 
 export function deactivate(): Thenable<void> | undefined {
-  if (!lifecycleController && !client) {
-    return undefined;
-  }
-  if (!lifecycleController) {
-    return client?.stop();
-  }
-  return lifecycleController.stop();
+  return lifecycleController?.stop();
 }


avifenesh:
Addressed Gemini review feedback in commit 2fc9fa0:
- simplified lifecycle stop delegate to avoid duplicated client-state clearing
- simplified deactivate() to `return lifecycleController?.stop();`

Re-ran `cd editors/vscode && npm run pretest && npm run test:unit` (45 passing).

PR #511, opened 2026-02-19
fix(core): normalize CRLF line endings at pipeline entry points (#480)

### LINKED ISSUE #480: Normalize CRLF line endings in frontmatter parser
## Audit Finding (Backend, Medium)

**Location**: `crates/agnix-core/src/parsers/frontmatter.rs:80`

The frontmatter parser handles CRLF files but leaves trailing `\r` in extracted frontmatter (e.g., `"\r\nname: test\r"`). While serde_yaml tolerates this, validators doing line-based analysis on raw frontmatter may encounter issues.

**Fix**: Normalize line endings when extracting frontmatter, or explicitly document CRLF behavior. Consider `frontmatter.replace("\r\n", "\n")`.

Found by `/audit-project`

avifenesh:
## Summary

- Adds `normalize_line_endings()` utility in `parsers/frontmatter.rs` - single-pass char scan, zero-allocation fast path for LF-only files via `Cow<str>`
- Normalizes CRLF and lone-CR to LF at all pipeline entry points: `validate_file_with_type`, `validate_content`, and `run_project_level_checks`
- Normalizes in the fix engine (`apply_fixes_with_fs_options`) so Fix byte offsets align with the content validators see
- Documents the LF-normalization contract on `FixResult` and `apply_fixes_with_fs_options`

## Why pipeline-level normalization

`split_frontmatter()` returns byte offsets (`frontmatter_start`, `body_start`) that callers add to local positions to compute absolute diagnostic positions. Normalizing only the extracted strings while keeping offsets referencing original CRLF content would break this arithmetic. Normalizing at the two pipeline entry points and the fix engine means all 27 validators and all offset calculations work on consistent LF content without changes.

## Test Plan

- [x] `normalize_line_endings` unit tests: LF-only (Borrowed), CRLF (Owned), lone CR, mixed, empty string
- [x] Proptest: output never contains `\r` for arbitrary input
- [x] `crlf_content_produces_same_diagnostics_as_lf` - in-memory path
- [x] `crlf_file_on_disk_produces_same_diagnostics_as_lf` - on-disk path via `validate_file()`
- [x] `lone_cr_content_produces_same_diagnostics_as_lf`
- [x] `crlf_validation_is_idempotent`
- [x] MockFS CRLF normalization test (dry-run and actual-write variants)
- [x] 2768 existing tests continue to pass

## Related Issues

Closes #480

### DIFF
--- a/crates/agnix-core/src/fixes.rs
+++ b/crates/agnix-core/src/fixes.rs
@@ -118,7 +118,7 @@ pub fn apply_fixes_with_options(
 ///
 /// File content is CRLF-normalized before fixes are applied, so byte offsets in
 /// [`Fix`] objects must reference LF-normalized positions (as produced by
-/// `validate_content` or `validate_file_with_type`). Files with CRLF endings
+/// `validate_content` or `validate_file`). Files with CRLF endings
 /// will be written back with LF endings.
 pub fn apply_fixes_with_fs_options(
     diagnostics: &[Diagnostic],
--- a/crates/agnix-core/src/parsers/markdown.rs
+++ b/crates/agnix-core/src/parsers/markdown.rs
@@ -122,18 +122,17 @@ fn extract_xml_tags_inner(content: &str) -> Vec<XmlTag> {
 ///
 /// # Security
 ///
-/// This function is NOT subject to `MAX_REGEX_INPUT_SIZE` limits because it uses
-/// pulldown-cmark's parser instead of regex. The limit only applies to regex-based
-/// extraction (`extract_xml_tags`).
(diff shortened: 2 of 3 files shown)


avifenesh:
Good catch. Fixed in 5866df8: replaced the silent strip with `out.push(' ')` so C0 control characters become spaces (1 byte → 1 byte) rather than being removed. This preserves byte-position alignment between the sanitized string and the CRLF-normalized content the fix engine operates on.

For VT (`\x0b`) and FF (`\x0c`) which are already converted to `\n` (also 1 byte → 1 byte), no change was needed there.

PR #548, opened 2026-02-21
feat: add Claude Code plugin wrapper

avifenesh:
Adds the agnix Claude Code plugin (agents, skills, commands) that was previously in the agentsys monorepo under plugins/agnix/.


avifenesh:
Holding off — need to decide how agnix plugin wrapper integrates with the existing repo.

PR #830, opened 2026-04-26
fix: add missing allowed-tools to plugin/skills/agnix/SKILL.md [skip changelog]

xiaolai:
> **Automated audit**: This PR was generated by [NLPM](https://github.com/xiaolai/nlpm-for-claude), a natural language programming linter, running via [claude-code-action](https://github.com/anthropics/claude-code-action). Please evaluate the diff on its merits.

## Bug

The two distributions of the agnix skill have diverged:

- `skills/agnix/SKILL.md` (root) declares `allowed-tools: Bash(agnix:*), Bash(cargo:*), Read, Glob, Grep`
- `plugin/skills/agnix/SKILL.md` (plugin) has no `allowed-tools` field at all

Without `allowed-tools`, the plugin skill runs with whatever permissions the invoking agent has — which may be broader than intended. The root skill explicitly restricts execution to `agnix` and `cargo` Bash commands plus read-only file operations.

**Impact**: Users who install agnix via the Claude Code plugin (`claude plugin install agnix`) get a skill that can execute arbitrary Bash commands, while users who load the root skill file get a properly scoped version. This is a security-relevant divergence.

## Fix

Added `allowed-tools: Bash(agnix:*), Bash(cargo:*), Read, Glob, Grep` to `plugin/skills/agnix/SKILL.md`, matching the existing declaration in `skills/agnix/SKILL.md`.

### DIFF
--- a/plugin/skills/agnix/SKILL.md
+++ b/plugin/skills/agnix/SKILL.md
@@ -2,6 +2,7 @@
 name: agnix
 description: "Use when user asks to 'lint agent configs', 'validate skills', 'check CLAUDE.md', 'validate hooks', 'lint MCP'. Validates agent configuration files against 405 rules across 10+ AI tools."
 argument-hint: "[path] [--fix] [--strict] [--target=claude-code|cursor|codex]"
+allowed-tools: Bash(agnix:*), Bash(cargo:*), Read, Glob, Grep
 ---
 
 # agnix


plugin/skills/agnix/SKILL.md
```diff
@@ -2,6 +2,7 @@
 name: agnix
 description: "Use when user asks to 'lint agent configs', 'validate skills', 'check CLAUDE.md', 'validate hooks', 'lint MCP'. Validates agent configuration files against 405 rules across 10+ AI tools."
 argument-hint: "[path] [--fix] [--strict] [--target=claude-code|cursor|codex]"
+allowed-tools: Bash(agnix:*), Bash(cargo:*), Read, Glob, Grep
```
avifenesh, line 5: It might also need npm in case the user wants to install from npm instead of cargo, and write if the user wants a report on a doc.
Other than that, LGTM
Thanks!

avifenesh:
Thanks for the audit - the issue you flagged is real. Closing in favor of the consolidated structural fix in #840, which addresses the root cause:

- Extends `scripts/sync-rule-bookkeeping.js` `COUNT_FILES` to cover `knowledge-base/README.md`, `knowledge-base/INDEX.md`, `plugin/commands/agnix.md`, `plugin/skills/agnix/SKILL.md`, `skills/agnix/SKILL.md`, and `plugin/.claude-plugin/plugin.json` so the rule-count drift you caught cannot recur.
- Syncs `plugin/.claude-plugin/plugin.json` + `npm/package.json` `version` fields from `Cargo.toml [workspace.package] version` so the version drift you caught also cannot recur.
- Wires `--check` into both `.github/workflows/ci.yml` (already there) and `scripts/pre-push-rust` (new), so both remote and local gates refuse to ship a drifted tree.
- One-time sweep applied all current drift (405 → 415 rules, 0.16.1 → 0.23.0 plugin, 0.20.0 → 0.23.0 npm, + the `allowed-tools` fix from #830).

If #840 merges without applying one of your findings correctly, please reopen. Sorry for the delay in getting back to the automated audits.

PR #878, opened 2026-05-06
feat(claude-code): triage v2.1.128 - add MCP-026 and CC-SET-002

### LINKED ISSUE #869: Tool release: Claude Code v2.1.128 (was v2.1.126)
## Claude Code v2.1.128

**Tier**: `s` &nbsp;&nbsp; **Previous baseline**: `v2.1.126` &nbsp;&nbsp; **Published**: 2026-05-04T23:01:47Z
**Source**: https://github.com/anthropics/claude-code/releases/tag/v2.1.128
**Repository**: https://github.com/anthropics/claude-code

---

#### Agnix-relevant changes

- `--channels` now works with console (API key) authentication — console orgs with managed settings must set `channelsEnabled: true` to enable -- matches: `New or renamed top-level keys in settings.json / managed-settings.json`
- MCP: `workspace` is now a reserved server name — existing servers with that name will be skipped with a warning -- matches: `MCP server config schema changes (fields, auth shapes, placeholders)`

#### Already-covered changes

_None._

#### Rule candidates

- MCP-026: Reserved MCP Server Name (`workspace`)
- CC-SET-002: Unknown Managed Settings Key (add `channelsEnabled` to valid-key allowlist)

#### Irrelevant changes (not reviewed)

35 items: UI/display fixes, CLI flag extensions, model picker changes, OTEL env handling, bug fixes for piping/resume/vim/clipboard/progress/crashes, error message wording, and streaming output changes.

<details><summary>Full upstream release notes</summary>

## What's changed

- Bare `/color` (no args) now picks a random session color
- `/mcp` now shows the tool count for connected servers and flags servers that connected with 0 tools
- `--plugin-dir` now accepts `.zip` plugin archives in addition to directories
- `--channels` now works with console (API key) authentication — console orgs with managed settings must set `channelsEnabled: true` to enable
- Updated `/model` picker: collapsed duplicate Opus 4.7 entries, and current Opus now shows as "Opus" instead of "Opus 4.7"
- Subprocesses (Bash, hooks, MCP, LSP) no longer inherit `OTEL_*` environment variables, so OTEL-instrumented apps run via the Bash tool no longer pick up the CLI's own OTLP endpoint
- MCP: `workspace` is now a reserved server name — existing servers with that name will be skipped with a warning
- Reconnecting MCP servers no longer flood the conversation with full tool-name lists on every reconnect — re-announced tools are summarized by server prefix
- SDK hosts now receive a persistent `localSettings` suggestion for Bash permission prompts, so "Always allow" writes to `.claude/settings.local.json`
- `EnterWorktree` now creates the new branch from local HEAD as documented, instead of `origin/<default-branch>` — unpushed commits are no longer dropped
- Auto mode: when the classifier can't evaluate an action, the error now includes a hint (retry, `/compact`, or run with `--debug`)
- Fixed focus mode briefly dimming the previous response when submitting a new prompt
- Fixed stray "4;0;" desktop notification on every `/exit` in Kitty and other terminals that interpret OSC 9 as a notification
- Fixed Remote Control showing an empty "Opening your options…" message on rate limit instead of actionable upsell options
- Fixed drag-and-drop image upload hanging on "Pasting text…" when the image read fails
- Fixed crash loop when piping very large input (>10 MB) to `claude -p` via stdin
- Fixed long URLs not being individually clickable on every wrapped row in fullscreen mode
- Fixed `/plugin` Components panel showing "Marketplace 'inline' not found" for plugins loaded via `--plugin-dir`
- Fixed MCP tool results dropping images when the server returns both structured content and content blocks
- Fixed fenced code blocks inside list items carrying leading whitespace into the clipboard on copy-paste
- Fixed tab navigation in `/config` stranding focus — the tab header now stays focused so arrows and Esc keep working
- Fixed markdown link labels being lost on terminals without OSC 8 hyperlink support — links now render as `label (url)` instead of just the URL
- Fixed sessions on 1M-context models with a smaller autocompact window being falsely blocked with "Prompt is too long" before reaching the actual API limit
- Fixed parallel shell tool calls: a failing read-only command (grep, git diff, ls) no longer cancels sibling calls
- Fixed banner showing "with X effort" on models that don't support effort
- Fixed `/fast` on 3P providers fuzzy-matching to an unrelated skill instead of showing "not available"
- Fixed Bedrock default model resolving to `global.*` instead of the region-appropriate prefix
- Fixed vim mode: `Space` in NORMAL mode now moves the cursor right, matching standard vi/vim behavior
- Fixed terminal progress indicator (OSC 9;4) flickering off between tool calls — stays visible across the full turn
- Fixed `/rename` without args failing on resumed sessions whose last entry is a compact boundary
- Fixed stale "remote-control is active" status lines from prior sessions appearing after `--resume`/`--continue`
- Fixed stale `installed_plugins.json` entries pointing at deleted cache directories polluting PATH
- Fixed MCP stdio servers receiving corrupted arguments when `CLAUDE_CODE_SHELL_PREFIX` is set and an argument contains spaces or shell metacharacters
- Fixed sub-agent progress summaries missing the prompt cache (~3× `cache_creation` reduction)
- Fixed `/plugin update` never detecting new versions of npm-sourced plugins
- Fixed sub-agent summaries firing repeatedly while a sub-agent's transcript is static, capping worst-case token cost on idle sub-agents
- Headless `--output-format stream-json`: `init.plugin_errors` now includes `--plugin-dir` load failures in addition to dependency demotions

</details>

---

### Action required

1. Review the release notes for changes that may affect agnix validation rules.
2. Update [`crates/agnix-core/src/config.rs`](https://github.com/agent-sh/agnix/blob/main/crates/agnix-core/src/config.rs) (`ToolVersions` / `SpecRevisions`) if the new version changes a validated field.
3. Update [`knowledge-base/RESEARCH-TRACKING.md`](https://github.com/agent-sh/agnix/blob/main/knowledge-base/RESEARCH-TRACKING.md) "Last Reviewed" for Claude Code.
4. After triage, run the workflow with `update_baselines: true` (and optionally `tool: claude-code`) and copy the JSON snippet from the job summary into [`.github/tool-release-baselines.json`](https://github.com/agent-sh/agnix/blob/main/.github/tool-release-baselines.json).

---
*Auto-opened by `.github/workflows/tool-release-watch.yml`. Run: https://github.com/agent-sh/agnix/actions/runs/25367518492*

avifenesh:
## Summary
Claude Code v2.1.128 shipped two config-affecting items that required new rules, not just a baseline bump. This PR implements both:

- **MCP-026 (HIGH)**: Reserved MCP server name `workspace`.
- **CC-SET-002 (MEDIUM)**: Non-boolean `channelsEnabled` setting.

Plus the usual baseline/changelog/research-tracking updates.

Closes #869.

## Upstream source

Release notes: https://github.com/anthropics/claude-code/releases/tag/v2.1.128

Two config-affecting bullets:
> MCP: `workspace` is now a reserved server name — existing servers with that name will be skipped with a warning
> `--channels` now works with console (API key) authentication — console orgs with managed settings must set `channelsEnabled: true` to enable

The remaining 35 items are triaged as agnix-irrelevant (UI/display, CLI flags that do not touch settings.json, model picker, OTEL env at subprocess spawn, piping/resume/vim/clipboard/progress/crashes, error message wording, streaming output) - documented in the CHANGELOG entry.

### MCP-026: Reserved MCP Server Name (HIGH)

**Location**: `crates/agnix-core/src/rules/mcp.rs`

Walks the top-level keys of `mcpServers` against a `RESERVED_MCP_SERVER_NAMES` constant (currently `["workspace"]`). Case-sensitive, string-literal-safe, and emits one diagnostic per reserved occurrence without suppressing sibling servers.

Key refactor: introduced `collect_mcp_server_name_offsets(content) -> Vec<(String, usize)>` as a shared JSON-key walker. MCP-023 (duplicates) and MCP-026 (reserved names) now both consume it via different predicates - one JSON walk per file regardless of how many name-based rules exist. Future server-name rules can reuse the same iterator.

Severity HIGH because silent-skip-at-startup is a functional failure the user won't notice in editor output.

### CC-SET-002: Non-boolean channelsEnabled (MEDIUM)

**Location**: `crates/agnix-core/src/rules/claude_settings.rs`

Parses JSON, looks up top-level `channelsEnabled`, emits a warning on string/number/array/object values. Explicit `null` and `false` are fine (`null` is JSON-convention "field-absent", `false` is a documented opt-out).

Validates across all three Claude Code settings paths (`settings.json`, `settings.local.json`, `managed-settings.json`). While v2.1.128's release note is scoped to "console orgs with managed settings", a mis-typed value is still wrong wherever it lands - trying to infer org-type here is too clever.

Same footgun shape as MCP-025 (`alwaysLoad`). Severity MEDIUM to match MCP-025 and CC-SET-001.

## Tests

- **MCP-026**: 6 unit tests in `rules::mcp::tests`:
  - `test_mcp_026_flags_workspace_server_name` (positive + line/col assertion)
  - `test_mcp_026_ignores_non_reserved_names`
  - `test_mcp_026_is_case_sensitive` (`Workspace` is not reserved)
  - `test_mcp_026_ignores_workspace_mention_in_string_literal`
  - `test_mcp_026_fires_alongside_siblings`
  - `test_mcp_026_suggestion_mentions_claude_code`
- Plus `tests/fixtures/mcp/reserved-server-name.mcp.json` so the parity test picks up fixture coverage.

- **CC-SET-002**: 11 unit tests in `rules::claude_settings::tests`:
  - `test_channels_enabled_absent_is_fine` / `_true_is_fine` / `_false_is_fine` / `_null_is_not_flagged`
  - `test_channels_enabled_quoted_string_flags` / `_number_flags` / `_array_flags`
  - `test_channels_enabled_runs_on_managed_settings` (cross-path)
  - `test_channels_enabled_line_points_at_key`
  - `test_channels_enabled_can_be_disabled` (rule-id disable path)
  - `test_channels_enabled_and_pr_url_template_coexist`

## Bookkeeping

Ran `node scripts/sync-rule-bookkeeping.js`, which updated 25 derived locations (rule count 416 -> 418 across CLAUDE.md, AGENTS.md, README.md, knowledge-base README/INDEX, plugin.json, SKILL.md files, VALIDATION-RULES.md footer stats, website docs + generated rule pages).

Also noticed `knowledge-base/RESEARCH-TRACKING.md` Claude Code row was missing `CC-SET` from its Rule Prefix column - pre-existing gap that's now conspicuous with two CC-SET rules. Fixed in this PR.

## Validation

- `cargo test --workspace` - all 4,820+ tests pass (17 new: 6 MCP-026 + 11 CC-SET-002)
- `cargo clippy --workspace --all-targets -- -D warnings` - zero warnings
- `node scripts/sync-rule-bookkeeping.js --check` - clean
- `python3 -m json.tool .github/tool-release-baselines.json >/dev/null` - passes
- Rule parity test (`rule_parity.rs`) passes: new rules registered, evidence metadata validated, fixture coverage inferred from `tests/fixtures/mcp/`.

## Test plan
- [x] MCP-026 unit tests pass
- [x] CC-SET-002 unit tests pass
- [x] Full workspace test suite passes
- [x] `cargo clippy` clean
- [x] Bookkeeping sync clean
- [x] Fixture added for MCP-026
- [x] Both new rules registered in `RULE_IDS`, wired into `validate()`, and included in `test_all_mcp_rules_can_be_disabled`
- [x] `knowledge-base/rules.json` entries include full `evidence` metadata per CLAUDE.md rule
- [x] `VALIDATION-RULES.md` sections added for both rules
- [x] CHANGELOG entries under `### Added` (new rules) and `### Changed` (baseline)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


avifenesh:
@gemini-code-assist thanks - both points applied in commit dea98f8:

1. `describe_json_type` now covers every `serde_json::Value` variant explicitly (Bool → "boolean", Null → "null"). CC-SET-002's caller still filters Bool/Null before reaching the helper, but it's now self-documenting and safe to reuse.

2. Good catch on the redundant walk. The PR description overclaimed - `validate()` was calling `collect_mcp_server_name_offsets` and `compute_line_starts` inside each of MCP-023 and MCP-026, so a file with both enabled paid the scan twice. Hoisted both into `validate()`, gated behind `is_rule_enabled(MCP-023) || is_rule_enabled(MCP-026)` so files with neither pay nothing, and now pass the precomputed vec + line_starts as `&[(String, usize)]` / `&[usize]` slices. Also switched the dedup `HashSet` to hold `&str` borrowed from the offsets vec instead of owned `String` clones - one allocation saved per distinct server name.

@claude please review the updated PR. The automated claude-review action is hitting the known "Internal error: directory mismatch / No buffered inline comments" bug again - triggering a manual pass. Context: two new rules (MCP-026 reserved `workspace` server name, CC-SET-002 non-boolean `channelsEnabled`) plus this gemini-requested refactor. Issue #869.

avifenesh:
@claude thanks - all three items addressed in 404ce44:

1. **Misleading group comment** - split the comment block so `validate_capability_keys` (which walks `capabilities`, not `mcpServers`) and the MCP-023/026 shared-walk block each carry their own reasoning.

2. **DiagnosticLevel::Error assertion** - added to `test_mcp_026_flags_workspace_server_name`. Mirrors MCP-025's warning-level assertion so an accidental error→warning (or vice versa) flip is caught by the test.

3. **Escape-handling limitation** - documented on `read_json_string_literal`. Explained that it's not a regression (MCP-023 has the same limitation), why a full decoder isn't justified (the evasion vector is the user lying to their own linter), and pointed at `serde_json::Value::as_object()` as the escape hatch if a third-party-config use case ever lands.

PR #902, opened 2026-05-08
fix(hooks): CC-HK-008 false-positive on ~-prefixed hook paths

### LINKED ISSUE #894: Two false-positive classes when validating Claude Code settings.json hooks: tilde paths and timeout units
When agnix validates a Claude Code settings.json containing hook entries, two classes of false-positive diagnostics fire on configurations that work correctly. Both stem from how agnix interprets fields whose semantics are defined by Claude Code's hook protocol. On a settings.json with ~12 hook entries, this produces ~24 noise diagnostics (one error + one warning per hook) and crowds out any real findings.

Reproduction is trivial — any valid Claude Code settings.json with hooks declared via ~-prefixed paths and millisecond timeout values triggers both bugs.

### Bug 1 — Tilde paths reported as not-found
**Symptom:**
error: Script file not found at '~/.claude/hooks/dippy-hook.py' (resolved to '/Users/philip/~/.claude/hooks/dippy-hook.py')
The "resolved to" path concatenates the working directory with the literal ~/..., never expanding the ~ to $HOME. The file exists at $HOME/.claude/hooks/dippy-hook.py and Claude Code finds it at runtime — agnix is the only thing that fails the existence check.

**Why this matters:** ~-prefixed paths are the canonical form for global hook scripts in Claude Code's ~/.claude/settings.json. The Claude Code documentation uses ~ paths in every hook example. If agnix flags every conventional setup as "missing scripts", the signal-to-noise ratio is unrecoverable.

### Reproduction:

`{
  "hooks": {
    "PostToolUse": [{
      "matcher": "Edit",
      "hooks": [{
        "type": "command",
        "command": "~/.claude/hooks/some-hook.py",
        "timeout": 5000
      }]
    }]
  }
}
$ touch ~/.claude/hooks/some-hook.py
$ agnix --target claude-code path/to/settings.json
error: Script file not found at '~/.claude/hooks/some-hook.py' (resolved to '$CWD/~/.claude/hooks/some-hook.py')`
**Suggested fix:** expand ~ and ~user per OS conventions before the existence check. In Rust this is a one-liner with shellexpand or equivalent; or dirs::home_dir() + manual prefix substitution if you want to avoid the dep.

### Bug 2 — Timeout field treated as seconds when Claude Code documents milliseconds

**Symptom:**
warning: Command hook at hooks.PostToolUse[matcher=Edit].hooks[0] has timeout 5000s exceeding 600s default
The hook entry's timeout is 5000 — five thousand milliseconds, i.e., five seconds, well below any reasonable default. agnix prints it as 5000s (five thousand seconds = 83 minutes) and warns that it exceeds the 600s default.

**Reference:** Claude Code's hook protocol documents timeout in milliseconds. See the [hooks reference](https://docs.claude.com/en/docs/agents-and-tools/claude-code/hooks-reference) and the schema for the hook config block.

**Reproduction:** any hook entry with timeout ≥ 600000 (600s in ms) triggers the warning; in practice almost any non-trivial timeout value triggers it because the threshold is only ~10× the default.

`{
  "type": "command",
  "command": "~/.claude/hooks/some-hook.py",
  "timeout": 5000
}`

→ flagged as exceeding 600s, when 5000ms = 5s is well within reason.

**Suggested fix:** either (a) treat timeout as milliseconds for --target claude-code validations and adjust the comparison threshold, or (b) per-target-tool unit declarations. Option (a) is the minimal fix; option (b) is more general if other target tools use seconds.

### Combined effect
A typical ~/.claude/settings.json with a dozen hook entries produces ~24 noise diagnostics (one of each class per hook). Real findings get buried. We currently work around this with a wrapper script that filters known false-positive lines before surfacing diagnostics, but the wrapper is a stopgap — these are upstream bugs and the right fix is in agnix.

### Environment
agnix 0.24.0 (npm install on macOS arm64)
Validating Claude Code settings.json (~/.claude/settings.json), generic and --target claude-code modes both reproduce
Hooks reference: https://docs.claude.com/en/docs/agents-and-tools/claude-code/hooks-reference

Happy to send a PR for either or both.

**Thanks for building this — the rule coverage and auto-fix flow are both excellent; just want to get the Claude-Code hook validation tightened up so the signal stays clean.**

avifenesh:
## Summary

Fixes bug 1 from #894 (reported by @philipstancil). CC-HK-008 was reporting `Script file not found at '~/.claude/hooks/hook.py' (resolved to '/cwd/~/.claude/hooks/hook.py')` on every hook using the canonical `~/` prefix - which is what every Claude Code docs example uses.

Bug 2 (timeout units) is not a bug - verified against both the public docs and the Claude Code v2.1.136 binary itself. Will reply on the issue.

Closes #894.

## Bug 1: tilde paths are runtime-expanded, not literal

Claude Code passes hook commands straight to the shell, which expands a leading `~/` to `$HOME/...` at runtime. agnix was resolving the literal `~/...` string relative to the project dir.

### Fix

- **`resolve_script_path`** now calls a new `expand_tilde_prefix()` helper that replaces a leading `~/` (or `~\\` on Windows) with `dirs::home_dir()` before the existence check. A bare `~` also resolves to `$HOME`.
- **`has_unresolved_env_vars`** treats `~user/...` forms as unresolved - we don't resolve arbitrary user home directories via nss/passwd. CC-HK-008 skips them (same pattern as `$FOO` vars).
- **Bare `~` / `~/` on systems without a readable HOME** is also treated as unresolved, so CC-HK-008 skips rather than false-positive on exotic setups.

### What is preserved

- `$CLAUDE_PROJECT_DIR/hook.sh` still resolves against the project dir (regression test).
- `$FOO/hook.sh` still skips the existence check (regression test).
- Mid-string `~` (e.g. `/opt/~backup/hook.sh`) is left literal - the shell only special-cases a leading tilde and Claude Code passes commands verbatim (regression test).
- Relative paths without `~` still join under `project_dir`, not `$HOME` (regression test).

### Tests

- **8 unit tests** in `rules::hooks::helpers::tests` cover: `is_tilde_user_path` true/false cases, `expand_tilde_prefix` expansion, leaving `~user/...` alone, leaving mid-string tildes alone, the resolvable-tilde-not-flagged behavior, `~user` being flagged, `$FOO` regression, the fully-resolved path.
- **2 integration tests** in `rules::hooks::tests` - one reproduces the exact issue content (`~/.claude/hooks/dippy-hook.py` with a `PostToolUse/Edit` matcher) and asserts no diagnostic contains `/~/`; the other confirms `~user/` paths skip CC-HK-008.

## Bug 2: timeout units - not a bug

The issue claims Claude Code documents hook `timeout` in milliseconds. I re-verified against two sources before replying:

1. **Public docs** at code.claude.com/docs/en/hooks (the hooks reference table):
   > `timeout` | no | Seconds before canceling. Defaults: 600 for command, 30 for prompt, 60 for agent

2. **Claude Code v2.1.136 binary** (installed via the npm package). The bundled Zod schema describes `timeout` as `"Timeout in seconds for this specific command"` (and `"...prompt evaluation"` / `"...tool call"` / `"...agent execution (default 60)"` for the other three hook types).

So the `timeout: 5000` example in the issue is 5000 seconds (~83 minutes), not 5 seconds. CC-HK-010's "exceeds 600s default" warning is correct. The internal `SESSION_END_HOOK_TIMEOUT_MS_DEFAULT` the user might have found is a Node.js `setTimeout` call that multiplies the user-facing seconds value by 1000; it's not the user-facing unit.

I'll reply on the issue with the citation so the user has the primary sources.

## Validation

- `cargo test --workspace` - all tests pass (10 new: 8 + 2)
- `cargo clippy --workspace --all-targets -- -D warnings` - zero warnings
- `cargo fmt --check --all` - clean

## Test plan

- [x] `dirs::home_dir()` expansion gated behind `filesystem` feature, matching `rules/imports.rs`
- [x] `~user/...` form skipped (no nss lookup)
- [x] Mid-string `~` left literal
- [x] `$CLAUDE_PROJECT_DIR` / `$FOO` semantics preserved via regression tests
- [x] Integration test reproduces the exact issue content
- [x] CHANGELOG entry credits reporter and explains the bug-1-fixed / bug-2-not-a-bug split


avifenesh:
Thanks @gemini-code-assist - good catch. Both points addressed in 0b02037:

1. **Split** the CC-HK-008 fix into its own bullet; the previous edit accidentally folded it into the kiro.validators entry.
2. **Tightened** the CC-HK-008 description to focus on the code change. Dropped the bug-2 timeout-units paragraph from the changelog - that discussion belongs on the issue where I replied with the primary-source citation (the Claude Code v2.1.136 binary's Zod schema explicitly says seconds).

PR #915, opened 2026-05-14
feat: add [[overrides]] for per-file rule suppression (#909)

### LINKED ISSUE #909: RFC: per-file rule overrides via .agnix.toml [[overrides]]
## Summary

Request a way to disable specific rules on specific files without disabling them globally, via a TOML `[[overrides]]` array-of-tables in `.agnix.toml`.

## Motivation

Today `.agnix.toml` offers only:

- `[rules].disabled_rules = [...]` — globally disables a rule across **all** files.
- `[files].exclude = [...]` — excludes a file from **all** rules.

There is no middle ground. Real-world example from my `~/.claude/CLAUDE.md`:

```markdown
## 🔴 CRITICAL: Documentation Update Trigger

**If the user says "in the future we should..." or "next time..." followed by any instruction:**

  **Example triggers:**
- "In the future we should check X before doing Y"
- "Next time, make sure to check Y first"
```

This block fires:

- `CC-MEM-007` / `PE-003` — "Weak constraint 'should' in critical section"
- `CC-MEM-005` — "Generic instruction 'make sure to' - Claude already knows this"

…but the `should` / `make sure to` strings are inside **quoted examples of what the user will say** — they're match patterns, not directives to the model. The autofix replaced them with `must`, which silently broke the trigger detection because the user phrases things with `should`/`next time`, not `must`.

I want to keep CC-MEM-007 / PE-003 / CC-MEM-005 enabled everywhere else in the repo, just suppressed for this one file.

## Proposed Design

```toml
# Suppress trigger-phrase warnings only on CLAUDE.md
[[overrides]]
paths = ["CLAUDE.md"]
disabled_rules = ["CC-MEM-007", "PE-003", "CC-MEM-005"]

# Different rules suppressed for skill examples
[[overrides]]
paths = ["**/skill-examples/*.md", "docs/snippets/**/*.md"]
disabled_rules = ["PE-001", "PE-004"]

# Yet another set of files with their own carve-out
[[overrides]]
paths = ["templates/**/*.md"]
disabled_rules = ["CC-MEM-009"]
```

Multiple `[[overrides]]` blocks stack: each file is matched against every block's `paths`, and the union of every matching block's `disabled_rules` is suppressed for that file.

vtmocanu:
Closes #909

## What

Adds a top-level `[[overrides]]` array-of-tables to `.agnix.toml` for disabling specific rules on specific file paths, without disabling them globally. Solves the common case where a file legitimately documents trigger phrases or quoted examples that would otherwise trip rules like CC-MEM-005 / CC-MEM-007 / PE-* — e.g., a project-level `CLAUDE.md` that documents what user-trigger phrases Claude should listen for.

```toml
[[overrides]]
paths = ["CLAUDE.md", ".claude/CLAUDE.md"]
disabled_rules = ["CC-MEM-005", "CC-MEM-007"]
```

Implementation notes:

- New `OverrideConfig { paths: Vec<String>, disabled_rules: Vec<String> }` plus `Vec<OverrideConfig>` on `ConfigData`. Same glob matcher as `[files].exclude` (`require_literal_separator = true`), same rule-ID-prefix validation as `[rules].disabled_rules`.
- New `PerFileLintConfig<'a>` view that `Deref`s to `&LintConfig` and overrides only `is_rule_enabled`. The `Validator` trait gained a `validate_per_file(&PerFileLintConfig<'_>)` method; the old `validate(&LintConfig)` is now a default-impl bridge. Both methods have defaults so pre-existing external `Validator` impls keep compiling and running unchanged — they just don't honor `[[overrides]]` until they migrate to `validate_per_file`. Internal pipeline dispatch (4 sites) calls `validate_per_file` directly so the view is constructed once per file. ~48 internal validator impls migrated; their bodies are unchanged thanks to `Deref`.
- Override semantics are purely additive (union with global `[rules].disabled_rules`); they cannot re-enable a globally disabled rule or override a category gate.
- Symlink fix: when `root_dir` has been canonicalized but the file path retains a symlink prefix (common when `~/.claude` lives in a dotfile repo and is symlinked into `$HOME`), `for_path` retries with `std::fs::canonicalize(path)` so overrides still match. Non-symlinked paths pay no syscall cost. Caught during manual end-to-end testing on the original `~/.claude/CLAUDE.md` use case from this issue.

## Validation

- `cargo test --workspace`: 30+ targets, all 0-failed (includes 3738 agnix-core unit tests + 3 new integration tests in `crates/agnix-core/tests/overrides_integration.rs`).
- `cargo clippy --workspace --all-targets -- -D warnings`: clean.
- `cargo fmt --check`: clean.
- Manual end-to-end on real `~/.claude/CLAUDE.md` (symlinked from a dotfile repo): override correctly suppresses CC-MEM-005 and CC-MEM-007 on the configured path; non-listed rules still fire.
- Pre-PR multi-agent review (QA / coding-standards / code-logic / API-interface / docs): 3 blocking findings folded back into the feature commits, re-verified clean on the amended branch.

**Full disclosure**: I don't have any programming skills so I asked Claude Code to implement this and did my best to test it and not create AI slop. All tests pass locally and the override behavior was verified end-to-end against my real `~/.claude/CLAUDE.md` (symlinked from a dotfile repo into `$HOME`, which surfaced and resolved a path-canonicalization bug during testing).


vtmocanu:
Thanks for the review! Both findings addressed in two new commits on the branch:

- HIGH (`rule_filter.rs:232` → now `4098a51`): `for_path` now routes the symlink-fallback canonicalize through `self.runtime.fs.canonicalize(path)` instead of `std::fs::canonicalize`. `FileSystem::canonicalize` was already defined on the trait and implemented on both `RealFileSystem` and `MockFileSystem`, so no trait change was needed. LSP + mock-fs scenarios now work correctly.
- MEDIUM (`schema.rs:153` → now `1b72cd6`): extracted a private `validate_rule_ids(field, rule_ids, warnings)` helper. Both call sites (`[rules].disabled_rules` and `[[overrides]].disabled_rules`) now share the same implementation, so future changes to the unknown-rule warning message land in one place.

`cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo fmt --check` all clean post-fix. Manual e2e on a real symlinked `~/.claude/CLAUDE.md` still suppresses the override-listed rules as expected.

avifenesh:
@vtmocanu Hi :) first, thank you again for picking this up. This is exactly the kind of contribution I was hoping for on #909, and the amount of testing/context you added here is really appreciated.

I did a local review pass and most of the implementation looks good to me. There is one edge case I found that I think we should fix before merge: `[[overrides]]` currently works for the normal file validators, but it does not affect project-level rules like `AGM-006`, `XP-004/005/006`, and `VER-001`.

Easy repro:

```bash
tmp=$(mktemp -d)
printf '%s\n' '[[overrides]]' 'paths = ["**/AGENTS.md"]' 'disabled_rules = ["AGM-006"]' > "$tmp/.agnix.toml"
printf '%s\n' '# Root' > "$tmp/AGENTS.md"
mkdir -p "$tmp/nested"
printf '%s\n' '# Nested' > "$tmp/nested/AGENTS.md"
cargo run -q -p agnix-cli --bin agnix -- --format json "$tmp"
```

Expected: no `AGM-006`, because both `AGENTS.md` files match `**/AGENTS.md` and the override disables that rule for them.

Actual: `AGM-006` still appears on both files.

Why it happens: the normal validator path uses `config.for_path(path)` and then calls `validate_per_file(...)`, so the per-file disabled rules are visible. But project-level checks in `crates/agnix-core/src/rules/project_level.rs` still call the global `config.is_rule_enabled("AGM-006")` / `config.is_rule_enabled("XP-...")`, so they never see the per-file override layer.

I think the clean fix is probably one of these shapes:

1. For diagnostics that are attached to a concrete file, check `config.for_path(&that_file).is_rule_enabled(rule_id)` before pushing the diagnostic.
2. Or add a tiny helper in `project_level.rs`, something like `is_rule_enabled_for_file(config, path, rule_id)`, and use it for `AGM-006`, `XP-004`, `XP-005`, and `XP-006` emissions.

`VER-001` is a little different because it reports on `.agnix.toml` or the project root, so I would either make it honor an override on `.agnix.toml`, or explicitly document/test that project-root config diagnostics are global-only. My preference is to keep the feature intuitive: if the diagnostic has a report path, use `for_path(report_path)`.

A regression test for `AGM-006` should be enough to lock the main bug. If you want to be extra safe, add one XP case too.

Also, local gates are clean on my side:

```bash
cargo fmt --check
cargo test -p agnix-core overrides
cargo test -p agnix-core --test overrides_integration
cargo test -p agnix-core for_path
cargo test --workspace
cargo clippy --workspace --all-targets -- -D warnings
```

So this is not a “bad PR” thing at all - it is a good PR with one project-level path that needs to be wired into the same override mechanism. Nice work getting the hard part in place.

vtmocanu:
Thank you @avifenesh for taking the time to review it, and provide such a thorough, I really apreciate it!
Will implement your findings and come back.

vtmocanu:
Back with the implementation. Two new commits on the branch:

**`5649aa2` — fix: honor `[[overrides]]` in project-level rules.**
Went with your option 1 (per-push-site `config.for_path(report_path).is_rule_enabled(rule_id)`) at all six emission sites in `crates/agnix-core/src/rules/project_level.rs`: AGM-006 inside the per-AGENTS.md loop, XP-004 read-error in the safe_read_file fallback, XP-004 build-conflict, XP-005 tool-conflict, XP-006 precedence-issue, and VER-001. Outer global `is_rule_enabled` guards retained as fast-paths — `[[overrides]]` is additive-only and can never re-enable a globally-disabled rule, so `if global { if per_file { push } }` is sound. VER-001 follows your stated preference: when `.agnix.toml` exists, an override on that path is honored; otherwise the report path falls back to `root_dir` and `for_path` handles both.

Three new integration tests in `crates/agnix-core/tests/overrides_integration.rs`:
- `overrides_suppress_agm006_on_matching_paths` mirrors your bash repro (root + `nested/AGENTS.md`, `paths = ["**/AGENTS.md"]`, expect 0 AGM-006).
- `overrides_suppress_xp004_build_conflict` covers the "extra safe" XP case (CLAUDE.md npm + AGENTS.md yarn, override on both, expect 0 XP-004).
- `overrides_suppress_ver001_on_agnix_toml` locks in the VER-001 shape (`.agnix.toml` present, no versions pinned, override targeting `.agnix.toml`, expect 0 VER-001).

Each test asserts the baseline (rule fires without override) before the suppression assertion, so a regression that drops both halves can't pass silently.

**`0636664` — docs: add CHANGELOG entry** under `[Unreleased]` so the `Verify Changelog` workflow stops blocking.

Verified your exact bash CLI repro end-to-end: baseline emits 2× AGM-006; with the `**/AGENTS.md` override applied, 0 AGM-006. Ran your verbatim local gate set (`cargo fmt --check`, `cargo test -p agnix-core overrides`, `cargo test -p agnix-core --test overrides_integration`, `cargo test -p agnix-core for_path`, `cargo test --workspace`, `cargo clippy --workspace --all-targets -- -D warnings`) — all clean. (Worth noting `cargo fmt --check` caught the new VER-001 test on the first pass; folded the rustfmt diff into `5649aa2` before commit.)

avifenesh:
@vtmocanu Thank you, this is a really good update. I pulled the new commits locally and verified the exact `AGM-006` repro from my previous comment is fixed:

```text
baseline_AGM006=2
with_override_AGM006=0
```

The new integration tests are also the right direction. I ran:

```bash
cargo fmt --check
cargo test -p agnix-core --test overrides_integration
cargo test -p agnix-core for_path
cargo test --workspace
cargo clippy --workspace --all-targets -- -D warnings
```

All clean.

I did find one more cross-file edge, and this one is partly on my earlier suggestion: push-site gating is enough when every participating file has the override, but for cross-file rules it is not enough when only one side is overridden. The disabled file still participates in the detector, and then suppression depends on which file the detector chooses as the diagnostic report path.

Repro:

```bash
tmp=$(mktemp -d)
printf '%s\n' '[[overrides]]' 'paths = ["CLAUDE.md"]' 'disabled_rules = ["XP-004"]' > "$tmp/.agnix.toml"
printf '%s\n' '# Claude' '' 'To install dependencies, run `npm install`.' > "$tmp/CLAUDE.md"
printf '%s\n' '# Agents' '' 'To install dependencies, run `yarn install`.' > "$tmp/AGENTS.md"
for i in $(seq 1 20); do
  cargo run -q -p agnix-cli --bin agnix -- --format json "$tmp" | grep -c '"rule": "XP-004"' || true
done | sort | uniq -c
```

On my machine this produced:

```text
11 0
 9 1
```

So the same config sometimes suppresses XP-004 and sometimes reports it. I think the cause is that XP-004 chooses `conflict.file1` from grouped `HashMap` iteration in the detector, so the report path can be either side of the conflict. With the current push-site gate, an override on `CLAUDE.md` only suppresses when `CLAUDE.md` happens to be `file1`.

I think the cleaner model for project-level rules is: if a file disables a project-level rule, that file should be removed from that rule's candidate set before the cross-file detector runs. Then the rule is deterministic and the semantics match the name `per-file override` better.

Suggested implementation shape:

- For `AGM-006`, build an `agents_md_paths_for_agm006` list containing only paths where `config.for_path(path).is_rule_enabled("AGM-006")`, and run hierarchy/multiple-file logic on that filtered list.
- For `XP-004`, build `file_contents_xp004` from only files where `config.for_path(path).is_rule_enabled("XP-004")`, then run `extract_build_commands` / `detect_build_conflicts` on that list.
- Same idea for `XP-005` and `XP-006`: filter the candidate content/layers by the rule before detection, not only before pushing the diagnostic.
- `VER-001` is fine as-is because it has a single report path rather than a set of participating files.

A good regression test would be the repro above but as a deterministic assertion: two files conflict, only `CLAUDE.md` disables `XP-004`, and the result should have zero `XP-004` diagnostics. That proves the disabled file no longer participates in the cross-file detector.

Again, the overall PR is in good shape. This is one of those annoying cross-file semantics edges, not a sign you did anything sloppy.

vtmocanu:
Thank you for the catch — the cross-file determinism flake is exactly what you described. New commit `74ea684` on the branch.

**What changed.** Switched from push-site gating (M20 shape) to your suggested filter-at-input model for the four cross-file rules. Each rule's participating set is now filtered before the detector runs:

- **AGM-006**: `agm006_paths` filtered from `agents_md_paths` by `for_path(p).is_rule_enabled("AGM-006")`. `len() > 1` and `check_agents_md_hierarchy` run on the filtered list. A filtered file is also invisible in other files' "other AGENTS.md files exist at:" listings — matching the "invisible to the rule" semantic you described.
- **XP-004 / XP-005 / XP-006**: each builds a per-rule filtered subset of `file_contents` before `extract_build_commands` / `detect_build_conflicts`, `extract_tool_constraints` / `detect_tool_conflicts`, and `categorize_layer` / `detect_precedence_issues` respectively.

Dropped the now-redundant push-site gates from these four rules. Kept them for **XP-004 read-error** (single-file emission, not cross-file) and **VER-001** (single report path) per your stated preference.

**Tests.** Two new integration tests in `tests/overrides_integration.rs`:

- `overrides_partial_xp004_deterministic_suppression` — your exact repro shape (CLAUDE.md `npm install` + AGENTS.md `yarn install`, override only on CLAUDE.md). Single deterministic assertion that XP-004 count == 0.
- `overrides_partial_agm006_no_cross_mention` — root + nested AGENTS.md, override only on nested. After filtering, the participating set drops below `len() > 1` so AGM-006 is fully suppressed; the root file would not mention the filtered nested file in its diagnostic either.

**Repro verification.** Ran your exact bash loop (extended to 50 iterations to be thorough):

```text
50 0
```

Uniformly zero across all 50 runs. The HashMap-order non-determinism is gone because the disabled file is no longer in the detector's input set.

**Local gates** (your verbatim list):

```bash
cargo fmt --check
cargo test -p agnix-core --test overrides_integration   # 8 pass
cargo test -p agnix-core for_path
cargo test --workspace                                  # 0 failed across 25 targets, 3738 agnix-core unit tests
cargo clippy --workspace --all-targets -- -D warnings
```

All clean.

CHANGELOG `Unreleased` entry reworded to describe the filter-at-input semantics; integration test count bumped 5 → 8.

avifenesh:
Thanks @vtmocanu, this round looks good.

I pushed one tiny cleanup on top: replaced a few em dashes in the new changelog/test prose with ` - ` to match our repo prose style. This is very much a personal issue with em dashes on our side, not a problem with the implementation.

Appreciate you iterating on this.

avifenesh:
@claude please review the latest head before I merge.

avifenesh (approved):
Looks good. Claude is also happy, checks are green, and the review threads are resolved. Thanks again for iterating on this carefully.

avifenesh:
@vtmocanu Merged. Thanks again for taking this on and for sticking with the review rounds. The final shape is much stronger because of that filter-at-input fix, and I really appreciate the care you put into the tests.

vtmocanu:
> @vtmocanu Merged. Thanks again for taking this on and for sticking with the review rounds. The final shape is much stronger because of that filter-at-input fix, and I really appreciate the care you put into the tests.

@avifenesh: Thank you for helping with the review rounds, I am happy that I was able to contribute. This has been the most thorough review I've received so far so thanks for the experience and congrats on doing a great job!

Also glad to see my feature request shipped so fast.

PR #967, opened 2026-05-24
feat(codex): validate managed requirements.toml (CDX-REQ-000/001) [skip changelog]

### LINKED ISSUE #965: Codex: add validator for requirements.toml (managed environment config)
Split out from the rust-v0.133.0 catch-up (#959 / #964).

Codex documented `requirements.toml` as a config surface in this release range (admin-only managed-environment config, e.g. `allow_managed_hooks_only = true`). It is listed in `tool-release-baselines.json` `codex.changes_of_interest.config_surfaces` but agnix has **no validator** for it - `grep -rn requirements.toml crates/agnix-core/src/` returns nothing.

Backing upstream code: `codex-rs/config/src/config_requirements.rs`. Note it is NOT part of `codex-rs/core/config.schema.json`, so the key catch-up did not cover it.

Scope: decide whether requirements.toml warrants a dedicated validator / schema, and if so add it (plus rules + tests + VALIDATION-RULES.md/rules.json entries as needed).

avifenesh:
## Summary
Closes #965. Adds the first validator for Codex's admin-written managed `requirements.toml` (system location: `/etc/codex/requirements.toml` on Unix, `%ProgramData%\OpenAI\Codex\requirements.toml` on Windows) - a surface agnix had no coverage for.

**Why an unknown-key check is the high-ROI wedge:** upstream `ConfigRequirementsToml` (`codex-rs/config/src/config_requirements.rs` @ `rust-v0.133.0`) has **no** `#[serde(deny_unknown_fields)]`. Codex silently drops a typo'd top-level key, so the intended managed constraint is never enforced and there is no other signal. The linter is the only catch - which is also why CDX-REQ-001 is a warning, not info (same rationale as CDX-004 for config.toml).

### Rules
- **CDX-REQ-000** (HIGH): invalid TOML syntax.
- **CDX-REQ-001** (MEDIUM): unknown top-level key, checked against the 19 documented keys. Both the `features` rename and the `feature_requirements` alias are accepted; the network table is valid only as `experimental_network` (rename-only, bare `network` is not a valid on-disk key).

### Plumbing
- New `FileType::CodexRequirements`, detected for `requirements.toml` under a `codex`/`Codex` parent (case-insensitive). The project `.codex/` dir is deliberately **not** matched - Codex never reads requirements.toml there.
- New `CodexRequirementsValidator` (modeled on `codex_plugin`), registered in `MiscProvider`, gated on the existing `codex` rule toggle.

### Scope & follow-ups (deliberate MVP)
- This PR is **parse + unknown-key only**. Cross-field invariants (`allowed_sandbox_modes` must include `read-only`; `enforce_residency` enum = `"us"`; the `experimental_network.domains` vs legacy `allowed_domains` mutex; reserved `[permissions.filesystem]` name) are genuinely high-value (load-failing) but each is its own evidence chain - will file a follow-up.
- `requirements.toml` is a **system-wide** file, so `agnix .` in a project tree won't normally encounter it; users will pass the path explicitly. Accepted low-probability trade-off: an unrelated `requirements.toml` under a literal lowercase `codex/` dir would be classified here (detection is path-only by design; documented in `detection.rs`).

## Test plan
- [x] `cargo test -p agnix-core` - all pass (incl. rules.json↔VALIDATION-RULES parity, api_contract exhaustiveness, CLAUDE/AGENTS byte-parity)
- [x] 7 unit tests: parse error, known-keys-clean, `feature_requirements` alias accepted, unknown scalar flagged, renamed `[network]` table flagged with correct line, disabled-when-codex-off
- [x] Detection test (Unix `/etc/codex/`, Windows `OpenAI/Codex/`, `.codex/` excluded, bare file excluded)
- [x] CLI smoke test: typo'd `allowed_sandbox_mode` + `[netwrk]` flagged at correct lines; valid keys clean
- [x] `cargo clippy` clean, `cargo fmt --check` clean
- [x] Rule bookkeeping synced (`--check` clean): 425 rules, 43 validators

### DIFF
--- a/crates/agnix-core/src/rules/codex_requirements.rs
+++ b/crates/agnix-core/src/rules/codex_requirements.rs
@@ -1,0 +1,294 @@
+//! Codex CLI managed-requirements validation (CDX-REQ-000, CDX-REQ-001).
+//!
+//! Validates Codex's admin-written `requirements.toml` (system location:
+//! `/etc/codex/requirements.toml` on Unix, `%ProgramData%\OpenAI\Codex\
+//! requirements.toml` on Windows). The file is deserialized upstream by
+//! `ConfigRequirementsToml` (codex-rs/config/src/config_requirements.rs) and,
+//! crucially, has **no** `#[serde(deny_unknown_fields)]` - so a typo in a
+//! top-level key is silently ignored by Codex and the intended constraint is
+//! never enforced. CDX-REQ-001 surfaces those typos; CDX-REQ-000 reports
+//! invalid TOML syntax.
+use crate::{
+    config::PerFileLintConfig,
+    diagnostics::Diagnostic,
+    rules::{Validator, ValidatorMetadata},
+};
+use std::path::Path;
+
+const RULE_IDS: &[&str] = &["CDX-REQ-000", "CDX-REQ-001"];
+
+/// Known top-level keys of `requirements.toml`, sourced from
+/// `ConfigRequirementsToml` @ openai/codex rust-v0.133.0. Both `features`
+/// (serde rename) and `feature_requirements` (serde alias) are accepted on
+/// disk; the network table is only valid as `experimental_network` (serde
+/// rename - the bare `network` field name is not a valid TOML key).
+const KNOWN_REQUIREMENTS_KEYS: &[&str] = &[
+    "allow_managed_hooks_only",
+    "allowed_approval_policies",
+    "allowed_approvals_reviewers",
+    "allowed_permissions",
+    "allowed_sandbox_modes",
+    "allowed_web_search_modes",
+    "apps",
+    "computer_use",
+    "enforce_residency",
+    "experimental_network",
+    "feature_requirements",
+    "features",
+    "guardian_policy_config",
+    "hooks",
+    "mcp_servers",
+    "permissions",
+    "plugins",
+    "remote_sandbox_config",
+    "rules",
+];
+
+pub struct CodexRequirementsValidator;
+
+impl Validator for CodexRequirementsValidator {
+    fn metadata(&self) -> ValidatorMetadata {
+        ValidatorMetadata {
+            name: self.name(),
+            rule_ids: RULE_IDS,
+        }
+    }
+
+    fn validate_per_file(
+        &self,
+        path: &Path,
+        content: &str,
+        config: &PerFileLintConfig<'_>,
+    ) -> Vec<Diagnostic> {
+        let mut diagnostics = Vec::new();
+        if !config.rules().codex {
+            return diagnostics;
+        }
+
+        // CDX-REQ-000: TOML syntax. Parse as a top-level table (stable across
+        // toml crate versions, unlike `toml::Value` document parsing).
+        let table: toml::Table = match toml::from_str::<toml::Table>(content) {
+            Ok(table) => table,
+            Err(error) => {
+                let (line, column) = toml_error_location(content, &error);
+                if config.is_rule_enabled("CDX-REQ-000") {
+                    diagnostics.push(
+                        Diagnostic::error(
+                            path.to_path_buf(),
+                            line,
+                            column,
+                            "CDX-REQ-000",
+                            format!("Invalid TOML in requirements.toml: {}", error.message()),
+                        )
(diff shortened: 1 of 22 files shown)


avifenesh:
Addressed all review comments in eb140b6:
- **Sort diagnostics** (gemini): CDX-REQ-001 now sorted by `(line, column)` since `toml::Table` (BTreeMap) iterates lexically, not in file order.
- **Dotted-key assignments + inline comments** (gemini, Copilot): `find_toml_key_line` now matches `key.sub = ...` and strips trailing `# ...` before structural matching; added `unknown_dotted_key_flagged_with_line` + `diagnostics_sorted_by_line` tests.
- **`toml_error_location` doc** (Copilot): clarified that column `0` is only the no-span fallback.
- **Stale category tables** (Copilot): fixed Codex CLI row (60->62, 31 HIGH / 26 MEDIUM) and TOTAL (216 HIGH / 181 MEDIUM) in both `INDEX.md` and `VALIDATION-RULES.md` (the bookkeeping sync updates footers but not these tables; VALIDATION-RULES total was also stale at 423).

crates/agnix-core/src/rules/codex_requirements.rs
```diff
@@ -0,0 +1,329 @@
...
+            Err(error) => {
+                let (line, column) = toml_error_location(content, &error);
+                if config.is_rule_enabled("CDX-REQ-000") {
+                    diagnostics.push(
+                        Diagnostic::error(
+                            path.to_path_buf(),
+                            line,
+                            column,
+                            "CDX-REQ-000",
+                            format!("Invalid TOML in requirements.toml: {}", error.message()),
```
avifenesh, line 81: This diagnostic message and the suggestion on the next line are hardcoded English strings. The comparable `codex.rs` uses `t!("rules.cdx_000.message", error = ...)` and `t!("rules.cdx_000.suggestion")`, and `codex_plugin.rs` follows the same pattern (`t!("rules.cdx_pl_002.message", error = ...)` / `t!("rules.cdx_pl_002.suggestion")`).

The four hardcoded strings in this file that need `t!()`:
1. Line 81: `format!("Invalid TOML in requirements.toml: {}", ...)`  
2. Line 83: `"Fix the TOML syntax error."`
3. Lines 101-104: `format!("Unknown requirements.toml key ...", ...)`
4. Lines 106-108: `"Remove or rename the key ..."`

This also needs corresponding locale entries in all three crate-level locale dirs (`agnix-core/locales/`, `agnix-cli/locales/`, `agnix-lsp/locales/`) per the locale-completeness always-on check.



---

crates/agnix-core/src/rules/codex_requirements.rs
```diff
@@ -0,0 +1,329 @@
...
+        if config.is_rule_enabled("CDX-REQ-001") {
+            for key in table.keys() {
+                if !KNOWN_REQUIREMENTS_KEYS.contains(&key.as_str()) {
+                    diagnostics.push(
+                        Diagnostic::warning(
+                            path.to_path_buf(),
+                            find_toml_key_line(content, key).unwrap_or(1),
+                            0,
+                            "CDX-REQ-001",
+                            format!(
```
avifenesh, line 101: Second instance of the same issue — this CDX-REQ-001 message and suggestion also need `t!()`. See comment on line 81.



---

avifenesh (commented):
## Summary

The new `CodexRequirementsValidator` is well-structured: clean file-type detection, correct TOML parsing, good alphabetical key list, proper sort-by-line, solid test coverage. The plumbing (registry, api_contract, rules.json, VALIDATION-RULES.md) is consistently wired.

One finding below: the validator uses hardcoded English strings for user-facing diagnostics and suggestions instead of the `t!()` i18n macro that every other validator in this crate uses.

### t! macro coverage

The new `codex_requirements.rs` contains 4 user-facing strings (2 messages + 2 suggestions) as inline English, while `codex.rs` (same category) and `codex_plugin.rs` (same model pattern per PR description) both import `rust_i18n::t` and wrap every diagnostic message and suggestion in `t!()`. Per the repo's always-on check, new user-facing strings in `agnix-core/src/` must use `t!()`, not inline English — diagnostics and suggestions are explicitly in scope (the "do not flag" carve-out is only for log/debug output).



---

Issue #957, opened 2026-05-24
Several rules conflict with current Claude Code documentation for skills and agents


saxonovsa:
I'm using agnix v0.27.1 to validate a Claude Code configuration on Windows. Several rules produce false positives compared to the current official Claude Code documentation. Citations below.

1. AS-008: "Description must be 1-1024 characters"
Official Claude Code skills docs (https://code.claude.com/docs/en/skills) state:

description ... the combined description and when_to_use text is truncated at 1,536 characters in the skill listing to reduce context usage.

The official limit is 1,536 characters, not 1,024. Skills with descriptions of 1,025-1,535 chars are correctly listed by Claude Code. agnix flags these as errors.

2. CC-AG-007: tools field "expected a YAML list, not a comma-separated string"
Claude Code sub-agents official docs (https://code.claude.com/docs/en/sub-agents) show this exact format in ALL examples:

---
name: code-reviewer
description: Reviews code for quality and best practices
tools: Read, Glob, Grep
model: sonnet
---
The comma-separated string format is the canonical format for sub-agents in the Claude Code docs. YAML list also works, but agnix rejects the canonical format with a parse error.

3. AS-016 false-firing on allowed-tools as YAML list
Claude Code skills docs say:

allowed-tools ... Tools Claude can use without asking permission when this skill is active. Accepts a space-separated string or a YAML list.

Both formats are explicitly documented as valid. agnix rejects YAML list format with "Failed to parse SKILL.md: allowed-tools: invalid type: sequence, expected a string".

4. AS-009: "Description must not contain XML tags"
Claude Code skills docs make no mention of forbidding angle brackets in description. Placeholders like <name> or <X> are commonly used to indicate parameter slots in description text. agnix flags these as XML tag errors.

5. CC-SK-008: "Unknown tool 'PowerShell'"
Claude Code on Windows supports a PowerShell tool. The list of "Known tools" in agnix appears to be missing PowerShell.

6. CC-SK-017 / XP-SK-001: aliases as unknown / client-specific field
The aliases field is supported by Claude Code skills, but agnix treats it as unknown / non-standard.

Environment
agnix 0.27.1 (latest at time of report)
Windows 11
Claude Code 2.x
~70 personal skills, configuration scoped via tools = ["claude-code"]
Suggested action
Either rule revision (correct limits/formats per current Claude Code docs) or making them configurable per-Claude-Code-version. For now I'm disabling these rules via .agnix.toml disabled_rules.

[...]

PR #972, opened 2026-05-24
feat(skill): scope AS-008/AS-009 by owning client (#957)

avifenesh:
## Summary
First PR addressing #957 (external report: agnix rules conflict with current docs). Skill validation applied the agentskills.io baseline (plus a Claude-sourced rule) uniformly to every `SKILL.md`, causing false positives on Claude Code skills. This scopes the two divergent rules to the skill's **owning client**.

**Client resolution** (your chosen "path + target"): from the skill path (`.claude/skills/`, `.agents/skills/`, … via the existing `per_client_skill` detector), falling back to the configured `target`/`tools` (so `tools = ["claude-code"]` on a bare `skills/` path resolves to Claude). `Unknown` → generic agentskills.io baseline.

- **AS-008** (description length): **1024** for the agentskills.io baseline — *verified to match Codex, OpenCode, and Kiro* — and **1536** for Claude Code skills (Claude truncates at 1536). Diagnostic now carries the resolved `%{max}`.
- **AS-009** (angle brackets in description): now fires **only for Codex skills**. Codex's `quick_validate.py` rejects `<`/`>`, but agentskills.io and Claude Code impose no such restriction. Re-sourced in `rules.json` (stale `platform.claude.com` → openai/codex `quick_validate.py`) and re-scoped `applies_to` claude-code → codex.

## Decisions backing this
- **Verified all 4 skill specs**: Codex/OpenCode/Kiro all enforce 1024 + space-string baseline; only Claude diverges → "Claude-vs-generic" for limits is correct (not per-tool).
- **AS-009 reversed from "delete" to "scope to Codex"** after finding Codex's validator genuinely bans angle brackets.

## Test plan
- [x] `cargo test -p agnix-core` — all pass; new tests `test_as_008_claude_allows_1536`, `test_as_009_angle_brackets_codex_only`
- [x] CLI smoke: Claude skill w/ 1200-char desc → "No issues found"; Codex skill w/ `<xml>` → AS-009 fires; Claude skill w/ angle brackets → no AS-009
- [x] `cargo clippy` + `cargo fmt --check` clean; bookkeeping `--check` clean (rule count unchanged, 425)

## Follow-ups (not in this PR)
- **AS-016** (allowed-tools string-vs-list parse error) — next PR.
- **CC-SK-008 / CC-SK-017 / CC-AG-007** gating + refresh (PowerShell + ~20 tools, when_to_use/arguments) — next PR.
- **XML-001** also flags `<name>` placeholders in descriptions (the generic XML-balance validator scans the description) — same `<name>` complaint, different validator; needs its own scoping.
- agentskills.io spec commit-watch tracking.
- 4 `AS-*` rules (AS-007/010/012/015) are CC/platform-sourced, not agentskills.io — audit/reclassify.


avifenesh:
Addressed all 4 review comments in 4dfc33b:
- **AS-009 detection** (Copilot): now matches Codex's `quick_validate.py` exactly - flags **any** `<` or `>` (not just tag-shaped pairs), auto-fix strips all angle brackets. Added a lone-`<` regression test. Removed the now-unused tag regex.
- **Case-insensitive tool matching** (Copilot): `skill_client_from_tool_str` lowercases before matching.
- **Doc/impl mismatch** (Copilot): clarified that an ambiguous multi-tool `tools` array resolves to `Unknown` (doesn't fall through to `target`).
- **VALIDATION-RULES.md AS-009** (Copilot): detection/fix wording aligned with the implementation.

PR #974, opened 2026-05-24
fix(skill,agent): correct + scope CC skill/agent rules (#957)

avifenesh:
## Summary
Part of #957. Three Claude-specific rules carried stale data or fired on the wrong clients.

- **CC-SK-008** (unknown tool): `KNOWN_TOOLS` refreshed to the current Claude Code built-in set — adds **`PowerShell`** (the reported false positive) plus `Agent`, `Cron*`, `Team*`, `EnterWorktree`/`ExitWorktree`, `ScheduleWakeup`, the MCP-resource tools, `ToolSearch`, etc. (legacy names kept for tolerance). Now **fires only for Claude Code skills** — other clients have their own tool vocabularies.
- **CC-SK-017** (unknown frontmatter field): added the documented `when_to_use` and `arguments` fields; now **Claude-only** (other clients handled by the per-client skill validator). `aliases` stays flagged — not a documented Claude field (reporter was mistaken on that one).
- **CC-AG-007** (agent parse error): sub-agent `tools`/`disallowedTools` accept a comma/space-separated **string** (canonical `tools: Read, Glob, Grep`) or a YAML list.

## Test plan
- [x] `cargo test -p agnix-core` — all pass. Updated 6 CC-SK-008/017 tests to Claude paths + 2 CC-AG-007 string-tools tests; new tests for PowerShell-known + Claude-only gating, `when_to_use`/`arguments` + gating, comma-separated agent tools.
- [x] `cargo clippy` + `cargo fmt --check` clean; bookkeeping `--check` clean; parity tests pass.

## Note
The `tools` untagged deserializer means an exotic non-string/non-list value (e.g. an integer) reports CC-AG-007 at the frontmatter start rather than the exact field line; precise line reporting for genuine YAML syntax errors is unchanged.


avifenesh:
Addressed CI failure + all review comments:
- **CI** (`test_format_text_verbose_shows_rule`): the `unknown-tool` fixture's `CC-SK-008` is now Claude-scoped and doesn't fire on a bare fixture path, leaving only the 2-segment `AS-017` (which the 3-segment `(AS|CC)-\w+-\d+` regex doesn't match). Repointed the test at the broader `invalid/skills` dir, which still emits a multi-segment code.
- **Seq trim** (gemini, Copilot): the YAML-list branch now trims + drops empties like the string branch.
- **Assertion** (Copilot): `test_cc_ag_007_type_error` now asserts the error lands at the frontmatter start (line 2), not a trivial `>= 1`.
- **Grammar** (Copilot): "a CC-AG-007 error".

PR #975, opened 2026-05-24
feat(release-watch): track agentskills.io spec via commit-path (#957)

avifenesh:
## Summary
Part of #957 (final piece). agentskills.io — the source of agnix's `AS-*` rules — publishes **no GitHub releases or tags**, so the release watcher couldn't detect spec changes (which is how the `AS-*` rules drifted in the first place).

Adds a **`commit_repo` + `commit_path`** source type to `scripts/check-tool-releases.sh` that tracks the latest commit SHA touching a file path. New baseline entry watches `docs/specification.mdx` in `agentskills/agentskills`. Any change opens a per-tool issue to diff the spec against the `AS-*` rules — complementing the weekly `spec-drift.yml` check.

## Test plan
- [x] `bash -n` clean; baselines JSON valid
- [x] Dry-run (`TOOL_FILTER=agentskills-spec UPDATE_BASELINES=true`): correctly fetches the SHA and reports `[ok] 6868401b64f7 matches baseline` (no spurious issue)
- [x] Baselined at the current spec commit (`6868401b64f7`, 2026-05-16) so it won't fire until the spec actually changes

## Note
Like the other watcher source branches (github releases, html scrape), this path is network-dependent and isn't unit-tested; the dry-run is the verification. The existing `check-tool-releases.test.sh` continues to cover the pure triage helper.


avifenesh:
Addressed the commit-path robustness comment in 9fa2583: pass `path`/`per_page` as `gh api` fields (`-f`/`-F`) so the path is URL-encoded correctly, guard an empty/failed API response before `jq`, and use the `latest_version` short SHA consistently in the issue body. Re-ran the dry-run: still `[ok]` against the baseline.

avifenesh:
Follow-up: `-f`/`-F` made `gh api` default to POST (returning an object → `jq` error). Added `-X GET` so the fields become query params; dry-run is back to `[ok]`.

PR #976, opened 2026-05-24
fix(skill): scope the whole CC-SK-* family to Claude Code skills (#957)

avifenesh:
## Summary
Audit follow-up from #957. The Claude Code skill rules (`CC-SK-001..020`) fired on **every** `SKILL.md` regardless of owning tool — so a Codex/OpenCode/Cursor skill was judged against Claude's model values, tool vocabulary, frontmatter fields, hooks, and fork/argument semantics (false positives for those tools' users).

Gates all `CC-SK-*` at the dispatch via `SkillClient::applies_claude_skill_rules()` — **true for `ClaudeCode` and `Unknown`** (an unscoped `SKILL.md` that could be a Claude skill), **false for any other known client**. This subsumes the per-rule gating added for CC-SK-008/017 in the previous PR (the redundant internal client checks are removed).

`Unknown` is included so unscoped skills keep current behavior; suppression applies only to skills under a known non-Claude client dir (`.agents/skills/`, `.opencode/skills/`, …).

## Test plan
- [x] `cargo test -p agnix-core` — all pass (280 skill tests); new `test_cc_sk_family_suppressed_for_known_non_claude_client` (CC-SK-001 fires unscoped, suppressed for Codex)
- [x] `cargo test -p agnix-cli` — all pass
- [x] `cargo clippy` + `cargo fmt --check` clean; rule count unchanged


avifenesh:
Addressed both review comments in 502d6a4: (1) the CC-SK gate is now config-aware — `Unknown` collapses both unscoped and ambiguous-multi-tool, so for `Unknown` it re-checks the tools scope and applies CC-SK only when the scope is empty or names `claude-code` (tools=["codex","cursor"] now correctly suppresses CC-SK); added a regression test. (2) Corrected the CC-SK-008/017 CHANGELOG wording to "Claude Code and unscoped skills".

PR #977, opened 2026-05-24
fix(xml): skip frontmatter in XML-001 balance check (#957)

avifenesh:
## Summary
Audit follow-up from #957. The XML balance validator scanned the **whole file**, so `<name>`/`<X>`-style placeholders in a skill's frontmatter `description` (or any frontmatter value) tripped **XML-001 "Unclosed XML tag"** — the other half of the reporter's `<name>` complaint (the AS-009 half was fixed in PR #972).

Masks the YAML frontmatter region before extracting tags (each non-newline byte → space, newlines + byte length preserved), so body XML balance and tag line/column/byte offsets are unaffected. No-op for files without frontmatter. Applies to every frontmatter file type the validator runs on (Skill, Agent, Copilot, …).

## Test plan
- [x] `cargo test -p agnix-core` — all pass; new tests: frontmatter `<name>`/`<X>` not flagged; unclosed `<thinking>` in body still flags XML-001 at the offset-adjusted line
- [x] CLI smoke: the Claude skill with `<name>` in description + 1100-char body now reports "No issues found" (was XML-001 before)
- [x] `cargo clippy` + `cargo fmt --check` clean


avifenesh:
Addressed all 3 review comments in the latest push: (1) **HR false positive** (Copilot) - masking now requires a closed block that parses as a YAML *mapping* (`is_yaml_frontmatter`), so Markdown `---` horizontal rules aren't mistaken for frontmatter; added a regression test. (2)+(3) **efficiency / infallible from_utf8** (gemini, Copilot) - `mask_frontmatter` now `split_at(body_start)`, iterates only the frontmatter region, and pushes the body as-is (no whole-file pass, no `from_utf8`).

PR #978, opened 2026-05-24
fix(skill): scope Claude-specific AS-* rules; re-source AS-012 (#957)

avifenesh:
## Summary
Audit follow-up from #957. Audited every `AS-*` rule against the current agentskills.io spec (+ its `skills-ref` validator). Four cited `platform.claude.com` instead of the generic spec:

- **AS-007** (reserved names), **AS-010** ("Use when" trigger phrase), **AS-015** (8 MB upload limit) — genuinely **Claude/platform-specific** (absent from the agentskills.io spec). Now gated to Claude Code (and unscoped) skills; suppressed for known non-Claude clients.
- **AS-012** (500-line `SKILL.md` body) — the opposite: it **is** in the agentskills.io spec ("Keep your main `SKILL.md` under 500 lines"). Its evidence was wrong (`platform.claude.com` / `claude-code`-only) → corrected to `agentskills.io/specification`, `spec_revision: 1.0`; remains a **generic** rule for all clients (no validator gating).

## Test plan
- [x] `cargo test -p agnix-core` — all pass (283 skill tests); new test (AS-007/AS-010 suppressed for Codex, AS-007 still fires unscoped); updated cursor-target test (AS-010 now Claude-scoped, AS-008 as universal control)
- [x] `cargo clippy` + `cargo fmt --check` clean; bookkeeping `--check` clean; rule count unchanged
- [x] VALIDATION-RULES.md notes updated for all four


avifenesh:
Addressed CI + all review comments:
- **CI** (`test_target_kiro/cursor_disables_cc_rules`): these used AS-010 as the non-CC control, but AS-010 is now Claude-scoped (suppressed for kiro/cursor targets). Switched the fixture to a >1024 description so the generic **AS-008** fires as the control.
- **AS-010 auto-fix guard** (Copilot): now uses the per-client length max (1536 Claude / 1024 else) instead of a hard-coded 1024.
- **AS-007 detection wording** (Copilot): clarified the name is matched case-insensitively (lowercased).
- **AS-008 "Too Short" name** (gemini): the rule name is a pre-existing misnomer — AS-008 checks the full `1..=max` range (fires for too-short **and** too-long), so the 1100-char control correctly triggers it. Out of scope to rename here.

PR #980, opened 2026-05-24
docs(changelog): allowed-tools list lenient by design (not a follow-up)

avifenesh:
## Summary
The #973 CHANGELOG entry said enforcing string-only `allowed-tools` for the agentskills.io-generic baseline was "tracked as a follow-up." It's a deliberate **won't-fix**, not a pending item: the YAML-list form is a Claude-specific feature, the tools that follow agentskills.io (Codex/OpenCode/Kiro) don't constrain the field's shape, and an unscoped skill using a list is almost certainly a Claude skill — so there's no client to safely warn against without re-introducing a false positive. Reworded the note accordingly.

Doc-only; no rule or code change.


avifenesh:
Applied the readability split (separate sentences). Kept `agnix` lowercase - it's the project's own name and is lowercase throughout the codebase, README, and this changelog (incl. the header "all notable changes to agnix"), so capitalizing only here would be inconsistent.

PR #1028, opened 2026-06-07
perf: use mimalloc global allocator on Linux

avifenesh:
## Summary

The `agnix`, `agnix-lsp`, and `agnix-mcp` binaries now use [mimalloc](https://github.com/microsoft/mimalloc) as the global allocator on Linux.

Linting is allocation-heavy: per-file parse trees, diagnostic vectors, and parallel rayon workers all churn the allocator. mimalloc reduces allocator contention versus the default system allocator.

## Details

- Added `mimalloc = { version = "0.1", default-features = false }` to `[workspace.dependencies]`.
- Wired it into each binary crate via a `[target.'cfg(target_os = "linux")'.dependencies]` block, so the dep and the `#[global_allocator]` static are both gated to Linux. macOS and Windows builds are unaffected.

## Testing

- `cargo check --workspace` - clean
- `cargo clippy -p agnix-cli -p agnix-lsp -p agnix-mcp` - clean
- `cargo test -p agnix-cli -p agnix-lsp -p agnix-mcp` - pass

## Changelog

Added under `[Unreleased]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Changed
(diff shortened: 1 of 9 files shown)


avifenesh:
Addressed the review suggestion: inlined `mimalloc::MiMalloc` at the `#[global_allocator]` site in all three binaries (`agnix`, `agnix-lsp`, `agnix-mcp`) and dropped the redundant `use mimalloc::MiMalloc;` imports. `cargo check --workspace` is clean.

PR #1072, opened 2026-06-22
[codex] accept plugin-prefixed skill names

avifenesh:
## Summary

Fixes false positives for valid plugin-provided skills named with the standard `plugin:skill-name` form.

- Parse skill names as either bare `skill-name` or prefixed `plugin:skill-name`.
- Validate AS-004/AS-005/AS-006 per segment so existing bare-name constraints still apply.
- Compare AS-017 parent-directory matches against only the local `skill-name` segment.
- Update AS-004 docs and locale text to describe the segment-based contract.

## Source

Reported via today's Gmail thread from Alexandr Tulai about `<plugin>:<skill-name>` false positives.

## Validation

- `cargo fmt --all --check`
- `bash scripts/check-locale-sync.sh`
- `node scripts/sync-rule-bookkeeping.js --check`
- `cargo test -p agnix-cli --test docs_website_parity`
- `cargo test --workspace`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- Live reproduction: `name: enhance:code-review` under `code-review/SKILL.md` now reports `No issues found`

### DIFF
--- a/crates/agnix-core/src/rules/skill/mod.rs
+++ b/crates/agnix-core/src/rules/skill/mod.rs
@@ -8,7 +8,10 @@ use crate::{
     rules::per_client_skill::{SkillClient, claude_skill_rules_apply, resolve_skill_client},
     rules::{Validator, ValidatorMetadata},
     schemas::hooks::HooksSchema,
-    schemas::skill::{SkillSchema, VALID_EFFORT_LEVELS, VALID_SHELLS, is_valid_skill_model},
+    schemas::skill::{
+        SkillSchema, VALID_EFFORT_LEVELS, VALID_SHELLS, is_valid_skill_model,
+        parse_skill_name_parts,
+    },
     validation::is_valid_mcp_tool_format,
 };
 use regex::Regex;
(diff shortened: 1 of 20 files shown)


crates/agnix-core/src/rules/skill/mod.rs
```diff
@@ -488,14 +575,14 @@ impl<'a> ValidationContext<'a> {
...
-                if !fixed_name.is_empty() && name_re.is_match(&fixed_name) {
+                if !fixed_name.is_empty() && is_valid_skill_name(&fixed_name) {
                     if let Some((start, end)) = self.frontmatter_value_byte_range("name") {
                         // Determine if fix is safe: only case changes are safe
                         let has_structural_changes = name_trimmed.contains('_')
                             || name_trimmed.contains(' ')
                             || name_trimmed
                                 .chars()
-                                .any(|c| !c.is_ascii_alphanumeric() && c != '-');
+                                .any(|c| !c.is_ascii_alphanumeric() && c != '-' && c != ':');
```
avifenesh, line 585: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

The AS-004 auto-fix safety check now allows `:` to count as non-structural, but `convert_to_kebab_case` is applied per-segment in `fixed_skill_name` while `is_case_only` compares against `name_trimmed.to_lowercase()`. For a prefixed name like `Build-Apps:Frontend` the per-segment fix yields `build-apps:frontend`, which equals `name_trimmed.to_lowercase()`, so it is correctly classified as a safe case-only fix — good. Worth double-checking that a name whose only change is case but which also contains an empty-segment edge (e.g. trailing/leading `:`) cannot slip through as `is_case_only`; those parse to `None` and go through `convert_to_kebab_case(name)` which strips the `:`, so `is_case_only` will be false there. No bug, just confirm the intent that `:`-containing structural rewrites stay unsafe.

avifenesh (commented):

---

Reviewed the plugin-prefixed skill-name change (AS-004/005/006/017, schema, locales, docs). The segment-based parsing in `parse_skill_name_parts` and the per-segment validators look correct and consistent across the rule path and schema path; all 12 locale files were updated in lockstep, the VALIDATION-RULES.md detection now references the actual `is_valid_skill_name` (the two earlier gemini comments are stale against this HEAD), and tests cover both bare and prefixed cases including AS-017 directory matching. `is_none_or` is fine under the repo MSRV of 1.91. One minor consistency note inline.

PR #1168, opened 2026-07-04
perf: avoid O(n^2) prefix rescan in the @import byte scanner

HetCreep:
## Problem

In `scan_imports_in_text` (`crates/agnix-core/src/parsers/markdown.rs`), the preceding-char check for each `@` byte did `text[..i].chars().last()`, which re-iterates the entire prefix `[0..i]` to fetch the single preceding char. On an input with many `@` bytes in one non-code span this is O(n^2). Unlike the sibling regex extractors, `extract_imports` has no `MAX_REGEX_INPUT_SIZE` guard and byte-scans up to the 1 MiB file cap, so a pathological ~1 MiB `@`-dense span is ~10^12 byte ops.

## Fix

Replace `.chars().last()` with `.chars().next_back()`, which decodes the preceding char in O(1) from the end of the slice - identical result, no prefix rescan. Turns the loop from O(n^2) to O(n).

Added `test_extract_imports_after_multibyte_char` to lock correct decoding of a multibyte char immediately preceding `@`.

Found during an external audit of v0.37.2. No behavior change (both return the last char); purely a complexity fix.


crates/agnix-core/src/parsers/markdown.rs
```diff
@@ -1,1736 +1,1749 @@
...
-                "Import with extension should be detected"
-            );
-            prop_assert_eq!(
-                &imports[0].path,
-                &format!("{}/{}.{}", dir, file, ext)
-            );
-        }
-    }
-}
+//! Markdown parser for extracting @imports, links, and checking XML tags
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

The file appears to have been rewritten with CRLF line endings: `git diff --check 494694cab6901810d7b9b292ddc74a866c6cde2d..1b33e297804bb27d7681ac198f180462f62fadd8` reports trailing-whitespace errors starting on this line (and across the rest of the file). The CI workflow runs `cargo fmt --check --all`, so this formatting-only line-ending change will fail the format gate; please normalize the file back to LF and keep the actual scanner/test changes.

avifenesh (commented):

---

`git diff --check` reports whitespace errors for this PR due to CRLF line endings being introduced throughout the touched Rust file, which will cause the repository's `cargo fmt --check --all` CI gate to fail.

crates/agnix-core/src/parsers/markdown.rs
avifenesh: Thanks - valid. I normalized `crates/agnix-core/src/parsers/markdown.rs` back to LF, added the missing changelog entry, and re-ran `git diff --check main`, `cargo fmt --check --all`, and the focused markdown parser regression test.

avifenesh:

avifenesh (approved):
Thanks - validated the current branch, the review threads are addressed, CI is green, and the explicit Claude gate completed. Approving.

PR #1169, opened 2026-07-04
fix: KNOWN_RULE_PREFIXES missing 11 prefixes -> spurious unknown_rule warnings

HetCreep:
## Problem

`KNOWN_RULE_PREFIXES` (`crates/agnix-core/src/config/schema.rs`), used by `validate_rule_ids` to warn on unrecognized `disabled_rules` entries, had fallen behind the rule-ID namespace in `knowledge-base/rules.json`. It was missing 11 prefixes covering 53 shipped rules: `CC-OS-`, `KR-AG-`, `KR-HK-`, `KR-MCP-`, `KR-PW-`, `KR-SET-`, `RC-SK-`, `CR-SK-`, `CL-SK-`, `CP-SK-`, `CX-SK-`.

A user who legitimately disables e.g. `disabled_rules = ["KR-AG-001"]` or `["CC-OS-001"]` got a spurious `core.config.unknown_rule` warning claiming the ID is unrecognized - even though the rule exists and the suppression still works (`rule_filter` matches by `starts_with`, independent of the warning). The docs already describe the broader `KR-*` family; the code-side prefix list was the piece that had drifted.

## Fix

- Add the 11 missing prefixes to `KNOWN_RULE_PREFIXES`.
- Add a completeness test (`known_rule_prefixes_cover_all_shipped_rules`) that asserts every `agnix_rules::RULES_DATA` ID is covered by a prefix, so the list can no longer silently drift behind `rules.json`.

## Follow-up (not in this PR)

The root cause is two prefix sources of truth (this const + the generated `agnix_rules::TOOL_RULE_PREFIXES`). A future change could derive `KNOWN_RULE_PREFIXES` from the generated data entirely; the completeness test above is the minimal guard in the meantime.

Found during an external audit of v0.37.2.


crates/agnix-core/src/config/schema.rs
```diff
@@ -1,259 +1,293 @@
...
+    "GM-",
+    "PE-",
+    "VER-",
+    "ROO-",
+    "AMP-",
+    "WS-",
+    "WS-SK-",
+    "KIRO-",
+    "KR-SK-",
+    "KR-AG-",
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

`KNOWN_*` allow-lists are expected to stay alphabetized relative to their neighbors so future drift audits are reliable. This block appends new prefixes out of order: `KR-AG-` now follows `KR-SK-`, `CC-OS-` is separated from the existing `CC-*` entries, and the new `*-SK-` prefixes are `RC/CR/CL/CP/CX` rather than sorted. Please insert `CC-OS-` with the `CC-*` group (`CC-MEM-`, `CC-OS-`, `CC-PL-`...), order the Kiro prefixes as `KR-AG-`, `KR-HK-`, `KR-MCP-`, `KR-PW-`, `KR-SET-`, `KR-SK-`, and sort the new skill prefixes as `CL-SK-`, `CP-SK-`, `CR-SK-`, `CX-SK-`, `RC-SK-`.

avifenesh (commented):

crates/agnix-core/src/config/schema.rs
avifenesh: Thanks - valid. I moved `CC-OS-` into the `CC-*` group, sorted the new `KR-*` and `*-SK-` entries, normalized the CRLF rewrite noise, added the changelog entry, and re-ran `git diff --check main`, `cargo fmt --check --all`, and the prefix coverage regression.

avifenesh:

avifenesh:
Thanks - valid nit. I updated the PR description so it no longer says docs/CONFIGURATION.md was stale; the docs already describe KR-*, and the actual drift was the code-side prefix list.

avifenesh (dismissed):
Thanks - validated the prefix fix, updated the PR description nit, CI is green, and the explicit Claude gate completed. Approving.

avifenesh:
Thanks - I updated the branch against current main after #1168 merged and resolved the changelog to keep both Unreleased entries. Verified the focused prefix regression, rule bookkeeping, and git diff --check.\n\n@claude please review the updated head before merge.

avifenesh (approved):
Thanks - revalidated after merging current main into the branch. CI is green, explicit Claude run completed, and the changelog conflict keeps both entries. Approving updated head.

PR #1171, opened 2026-07-04
ci: attest SLSA build provenance for release binaries

HetCreep:
## What

Adds GitHub-native SLSA build-provenance attestation so consumers can verify a downloaded binary was built by this repo's CI (`gh attestation verify --repo agent-sh/agnix <file>`), as defense-in-depth above the SHA-256 checksums.

## Change (revised per review)

Provenance is generated in the **`build` matrix job**, immediately after each target's archives are created and **before `upload-artifact`**, with `id-token: write` + `attestations: write` scoped to the `build` job only. The `release`/publish job carries no attestation tokens, so a compromised publish step cannot mint a valid attestation for a swapped archive - each artifact is attested at build time, bound to the build.

## Current upstream references

- GitHub's current artifact-attestation docs show build provenance for binaries with `actions/attest@v4` and `subject-path`: https://docs.github.com/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds#generating-build-provenance-for-binaries
- The pinned `actions/attest@a1948c3f` README documents Provenance mode: when no `sbom-path` or predicate inputs are supplied, it auto-generates SLSA build provenance: https://github.com/actions/attest/blob/a1948c3f048ba23858d222213b7c278aabede763/README.md#attestation-modes
- The pinned `action.yml` marks `predicate-type`, `predicate`, and `predicate-path` as not required; those inputs are required only for custom attestations: https://github.com/actions/attest/blob/a1948c3f048ba23858d222213b7c278aabede763/action.yml
- `actions/attest-build-provenance@v4.1.1` documents itself as a wrapper over `actions/attest` and says new implementations should use `actions/attest`: https://github.com/actions/attest-build-provenance/tree/v4.1.1#usage
- Verified with `git ls-remote`: `actions/attest` tag `v4.1.1` resolves to `a1948c3f048ba23858d222213b7c278aabede763`.

Thanks @avifenesh for catching that the original attest-in-release-job version left the same self-signed-release gap open.

Found during an external audit of v0.37.2.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,7 +6,7 @@ on:
       - 'v*'
 
 permissions:
-  contents: write
+  contents: read
 
 env:
   CARGO_TERM_COLOR: always
@@ -141,7 +141,7 @@ jobs:
           }
 
       - name: Attest build provenance
-        uses: actions/attest-build-provenance@0f67c3f4856b2e3261c31976d6725780e5e4c373 # v4.1.1
+        uses: actions/attest@a1948c3f048ba23858d222213b7c278aabede763 # v4.1.1
         with:
           subject-path: |
             agnix-*${{ matrix.target }}.tar.gz
@@ -185,6 +185,8 @@ jobs:
     name: Create Release
     needs: [build, test]
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -174,6 +174,10 @@ jobs:
     name: Create Release
     needs: [build, test]
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      id-token: write
+      attestations: write
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
@@ -182,6 +186,13 @@ jobs:
         with:
           path: artifacts
           merge-multiple: true
+
+      - name: Attest build provenance
+        uses: actions/attest-build-provenance@0f67c3f4856b2e3261c31976d6725780e5e4c373 # v4.1.1
+        with:
+          subject-path: |
+            artifacts/*.tar.gz
+            artifacts/*.zip
 
       - name: Generate release notes
         id: notes
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,6 +19,10 @@ jobs:
   build:
     name: Build (${{ matrix.target }})
     runs-on: ${{ matrix.os }}
+    permissions:
+      contents: read
+      id-token: write
+      attestations: write
     strategy:
       fail-fast: false
       matrix:
@@ -136,6 +140,13 @@ jobs:
             (Get-FileHash agnix-mcp-${{ matrix.target }}.zip -Algorithm SHA256).Hash.ToLower() + "  agnix-mcp-${{ matrix.target }}.zip" | Out-File -Encoding ASCII agnix-mcp-${{ matrix.target }}.zip.sha256
           }
 
+      - name: Attest build provenance
+        uses: actions/attest-build-provenance@0f67c3f4856b2e3261c31976d6725780e5e4c373 # v4.1.1
+        with:
+          subject-path: |
+            agnix-*${{ matrix.target }}.tar.gz
+            agnix-*${{ matrix.target }}.zip
+
       - name: Upload artifact
         uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
         with:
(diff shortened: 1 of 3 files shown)


HetCreep:
Thanks - agreed that defaulting the workflow token to read-only is the right least-privilege posture. I have kept this PR scoped to just the provenance step, because tightening the top-level `contents: write` needs a per-job audit first: e.g. the `version-docs` job may commit generated docs back and would then need its own `contents: write`, and I could not verify each publish job blind without running the tag-triggered release. Happy to do that as a focused follow-up. Note the `release` job here already declares its own minimal scoped permissions.

.github/workflows/release.yml
```diff
@@ -1,623 +1,634 @@
...
+        run: cargo test --workspace
+
+  release:
+    name: Create Release
+    needs: [build, test]
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      id-token: write
+      attestations: write
```
avifenesh: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

These permissions are granted to the entire `release` job, which also contains the later release-upload/publish steps (`softprops/action-gh-release` and `gh api`). Because any subsequent step in this job can use the job's OIDC token and `attestations: write` token, a compromised release/upload step could still replace an archive and mint a valid attestation for the tampered file. That leaves the same self-signed release-step gap the PR is trying to close. Please isolate provenance generation into the build matrix before `upload-artifact` (or a separate non-publishing attestation job) and keep the release publishing job without `id-token`/`attestations: write`.

avifenesh (commented):

---

Found one issue with the provenance boundary that affects the stated release-step threat model.

HetCreep:
You're right - thanks, that's a real gap. Attesting inside the `release` job means a compromised publish step (`softprops/action-gh-release` / `gh api`) could swap an archive and still mint a valid attestation for it, leaving exactly the self-signed-release hole this was meant to close.

Fixed: moved provenance generation into the `build` matrix job, right after each target's archives are created and before `upload-artifact`, with `id-token: write` + `attestations: write` scoped to the `build` job only. The `release`/publish job no longer carries those permissions. Each artifact is now attested at build time, bound to the build, before any downstream step can touch it. Pushed.

.github/workflows/release.yml
```diff
@@ -1,623 +1,634 @@
...
-          fi
+name: Release
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+permissions:
+  contents: write
```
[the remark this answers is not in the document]
avifenesh: Thanks - valid. I changed the workflow default to `contents: read` and gave `contents: write` only to the release publishing job, with a regression test to keep that boundary in place.

.github/workflows/release.yml
avifenesh: Thanks - valid threat-model point. The attestation step stays in the build job, before artifacts are uploaded, and release publishing now has its own `contents: write` job permission without OIDC/attestation permissions. I also switched the new implementation to the current `actions/attest` v4.1.1 SHA and added release-workflow coverage.

avifenesh:

avifenesh:
Thanks - I checked this against the current GitHub artifact-attestation docs before changing anything. This concern looks stale now: GitHub's current docs for build provenance use actions/attest@v4 with subject-path, and the actions/attest-build-provenance README says v4 is now just a wrapper and new implementations should use actions/attest instead. I also verified the pin: actions/attest v4.1.1 resolves to a1948c3f048ba23858d222213b7c278aabede763.\n\nSo I am leaving the workflow on actions/attest@a1948c3f... rather than switching to the wrapper. @claude please re-check this attestation-action concern against the current GitHub docs before merge.

avifenesh:
Thanks for re-checking. Confirmed with the current upstream sources now:\n\n- GitHub's artifact attestation docs show build provenance for binaries with actions/attest@v4 and subject-path: https://docs.github.com/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds#generating-build-provenance-for-binaries\n- The pinned actions/attest@a1948c3f README documents the attestation modes: no sbom-path and no predicate inputs means Provenance mode, which auto-generates SLSA build provenance: https://github.com/actions/attest/blob/a1948c3f048ba23858d222213b7c278aabede763/README.md#attestation-modes\n- The pinned action.yml marks predicate-type/predicate/predicate-path as required: false, and the text says predicate-type is required only when using predicate or predicate-path for custom attestations: https://github.com/actions/attest/blob/a1948c3f048ba23858d222213b7c278aabede763/action.yml\n- actions/attest-build-provenance@v4.1.1 documents itself as a wrapper over actions/attest and says new implementations should use actions/attest: https://github.com/actions/attest-build-provenance/tree/v4.1.1#usage\n- I also verified the pin with git ls-remote: actions/attest v4.1.1 resolves to a1948c3f048ba23858d222213b7c278aabede763.\n\nGiven that, I disagree with switching this PR to the wrapper action. The current actions/attest@a1948c3f step with subject-path is the current documented build-provenance path.

avifenesh (dismissed):
Thanks - validated the provenance-boundary fix, CI is green, and the attestation-action concern is addressed with current upstream docs and pin evidence in the PR. Approving.

avifenesh:
Thanks - I updated the branch against current main after #1168/#1169/#1170 merged and resolved the changelog to keep the Added provenance entry plus all Fixed entries. Verified cargo test --test release_workflow release_workflow_scopes_attestation_and_release_permissions, cargo fmt --check --all, and git diff --check.\n\n@claude please review the updated head before merge.

avifenesh (approved):
Thanks - revalidated after merging current main into the branch. CI is green, explicit Claude run completed, and the changelog keeps all entries plus the provenance Added item. Approving updated head.

PR #1175, opened 2026-07-04
fix: harden security followups

### LINKED ISSUE #1149: [BUG] frontmatter.rs doc comment overclaims serde_norway's memory/nesting protection
## Description

The module doc comment in `crates/agnix-core/src/parsers/frontmatter.rs` (around lines 11-14) states that the YAML parser "preserves ... internal protections against excessive memory usage and stack overflow from deeply nested structures." `serde_norway` does not provide this. It is a hard-fork of dtolnay's `serde-yaml`, which has no billion-laughs / alias-expansion / nesting-depth limit built in (the well-known upstream gap tracked as serde-yaml issue #53; hardened forks such as `serde-yaml-bw` advertise billion-laughs protection precisely because the base does not have it).

The actual protection against deeply nested / expansion-bomb input is agnix's own `check_yaml_depth` pre-parse guard plus the 1 MiB size cap - which the comment lists as a separate, additional measure. Crediting the library with a safety property it lacks is a documentation-rot risk: a future maintainer could weaken or remove `check_yaml_depth` believing the parser is a backstop, silently removing the only real defense.

## Evidence

- `serde_norway` is a hard-fork of `serde-yaml`; its published crate carries no expansion/nesting limit (verified against the crate docs and the upstream serde-yaml issue #53, 2026-07-04).
- The real guard is `check_yaml_depth` (`MAX_YAML_DEPTH`) + the 1 MiB read cap, applied by agnix before/around the parse call - independent of the library.

## Expected Behavior

The comment describes the parser's actual guarantees. Memory/nesting protection is attributed to agnix's own `check_yaml_depth` + size cap, not to `serde_norway`.

## Actual Behavior

The comment attributes billion-laughs / stack-overflow protection to `serde_norway`, which does not provide it.

## Environment

- agnix version: 0.37.2
- OS: n/a (documentation/source review)

### LINKED ISSUE #1150: [BUG] Panic isolation is incomplete: project-level checks run outside catch_unwind
## Description

The panic-to-diagnostic isolation added around per-file validation only wraps the per-file path. In `validate_project_with_registry` (`crates/agnix-core/src/pipeline.rs`), the `catch_unwind` at ~line 778 wraps `validate_file_with_type` for each file, converting a per-file validator panic into a `file::panic` diagnostic instead of aborting the run. But `run_project_level_checks` (~line 859 - AGM-006, XP-004/005/006, VER-001) runs after the parallel reduce, outside any `catch_unwind`. A panic in project-level logic therefore aborts the entire `validate_project`, which is the exact failure mode the per-file isolation was added to prevent.

This is a structural gap, reported honestly as such: production `crates/agnix-core/src/rules/project_level.rs` (lines 1-306, before the `#[cfg(test)]` module) contains no `unwrap`/`expect`/raw content-slicing, and the markdown sub-parsers it calls (`extract_imports_inner` / `extract_xml_tags_inner` / `extract_markdown_links_inner` in `parsers/markdown.rs`) already carry their own `catch_unwind`. No panic-triggering input was demonstrated. The point is that the panic surface is covered asymmetrically - the guarantee "one bad file cannot abort the whole lint run" does not extend to the project-level rules, which also process untrusted file paths and content.

## Evidence

- `crates/agnix-core/src/pipeline.rs:778` - `catch_unwind(AssertUnwindSafe(|| validate_file_with_type(...)))` (per-file, guarded).
- `crates/agnix-core/src/pipeline.rs:859` - `run_project_level_checks(...)` called directly, no surrounding `catch_unwind`.
- Grep of all production `catch_unwind` sites: only `pipeline.rs:778` and the three `parsers/markdown.rs` parsers; none wrap the project-level section.

## Expected Behavior

A panic in a project-level check is contained the same way a per-file panic is (converted to a diagnostic), so no single input can abort the whole validation run.

## Actual Behavior

A panic in `run_project_level_checks` propagates and aborts `validate_project`.

## Suggested Fix

Wrap the project-level check block in the same `catch_unwind` + `panic_payload_message` pattern used for per-file validation, emitting a diagnostic on panic. Optionally add a `compile_error!`/CI assertion tying the isolation feature to `panic = "unwind"` so re-adding `panic = "abort"` to `[profile.release]` cannot silently disable it.

## Environment

- agnix version: 0.37.2
- OS: n/a (source/structural review)

### LINKED ISSUE #1154: [BUG] download.sh checksum parse ignores the filename; verifiers diverge on artifact binding
## Description

The bash installer parses the checksum sidecar with `awk 'NF {print $1; exit}'` (`scripts/download.sh:155`), which takes the first non-empty line's first field (the hash) and ignores the filename entirely. It then validates only the hash **format** (`grep -Eq '^[0-9a-f]{64}$'`), never that the hash actually corresponds to the downloaded artifact. The TypeScript (`editors/vscode/src/checksum.ts`) and Kotlin (`AgnixBinaryDownloader.kt`) verifiers do check the filename field (`parts[1]`), so the three verifiers enforce different contracts on the same sidecar format.

This fails closed - a wrong hash produces a mismatch and the script aborts - and works today only because the release convention ships a single-artifact `.sha256` per binary. If a combined `SHA256SUMS`-style file (multiple `<hash>  <file>` lines) were ever served, `download.sh` would pick the first line's hash regardless of which artifact it belongs to and compare the agnix binary against the wrong file's hash. It is fragile and inconsistent with the other two verifiers, not currently exploitable.

## Evidence

- `scripts/download.sh:155` - `EXPECTED_SHA="$(awk 'NF {print $1; exit}' "${CHECKSUM_FILE}" | tr '[:upper:]' '[:lower:]')"` then `grep -Eq '^[0-9a-f]{64}$'` (format only).
- `editors/vscode/src/checksum.ts` and JetBrains `AgnixBinaryDownloader.kt` both parse and check the filename field (`parts[1]`) and reject a hash whose filename does not match the artifact.
- Reproduced by running the awk parse against a two-line `SHA256SUMS`-style file: it returns the first line's hash irrespective of filename.

## Expected Behavior

`download.sh` binds the selected hash to the downloaded artifact's filename (matching the TS/Kotlin verifiers), so a multi-entry checksum file cannot cause the wrong hash to be used.

## Actual Behavior

`download.sh` uses the first line's hash and never checks the filename; the three verifiers diverge.

## Environment

- agnix version: 0.37.2
- OS: n/a (installer script; reproduced via awk)

### LINKED ISSUE #1155: [BUG] MCP server returns canonicalized absolute host paths in diagnostics and errors (info disclosure to the client)
## Description

The MCP server resolves every input path to its canonicalized absolute filesystem path before validating, then returns that absolute path unmodified in both the error text and the `Diagnostic.file` field. A client that passes a relative path like `SKILL.md` gets back `C:\\Users\\<username>\\...\\SKILL.md` in every diagnostic and in any I/O error, exposing the server host's directory layout and OS username to whatever MCP client is calling the tool.

Unlike the CLI (where the invoking user already has full disk access), an MCP client is a different trust boundary - potentially a hosted or remote agent - so leaking the absolute host path is a real, if low-severity, information disclosure. This is a hardening gap under the "trusted input, local execution" threat model, not a confidentiality bypass of a secret.

## Evidence

- `crates/agnix-mcp/src/main.rs` `resolve_path_within_workspace` canonicalizes the candidate (main.rs:387-389 `candidate.canonicalize()`) and returns `canonical_candidate` (the absolute path) at main.rs:395.
- That `PathBuf` flows into `core_validate_file(&file_path, &config)` (main.rs:442) and the `Err` is formatted as `format!("Failed to validate file: {}", e)` (main.rs:443) where `FileError` Display embeds the absolute path.
- The same path flows unchanged into `Diagnostic.file`, then `DiagnosticOutput::from` does `d.file.display().to_string()` (main.rs:184) with no relativization against the caller's input.

## Suggested fix

Relativize the returned path against the caller-supplied `input.path` (or the workspace root) before embedding it in diagnostics/errors returned over MCP, so the client sees the path it asked about, not the server's absolute layout.

## Environment
- agnix version: 0.37.2

### LINKED ISSUE #1156: [BUG] MCP tool handlers run core validation with no panic isolation (a panicking validator kills the server loop)
## Description

The MCP tool handlers call core validation directly in the async rmcp `#[tool]` fn with no `catch_unwind` and no `spawn_blocking`, while the equivalent LSP path IS panic-isolated. A panic in any validator (or in the project-level checks) unwinds through the rmcp dispatch future and aborts the stdio connection task, killing the server for all subsequent requests. `panic = "abort"` is not set (default unwind), so this is an unwinding crash of the connection loop.

The LSP backend catches the `spawn_blocking` `JoinError` and converts a panic into an `agnix::internal-error` diagnostic, keeping the server alive - the MCP side has no equivalent guard. This is the "one bad request crashes the server loop" failure mode, defended in LSP but not MCP.

Related to (but distinct from) #1150: that issue is about `run_project_level_checks` running outside the per-file `catch_unwind`; this is about the MCP entry points having no isolation at all.

## Evidence

- `crates/agnix-mcp/src/main.rs:442` `let outcome = core_validate_file(&file_path, &config)...?;` and `:466` `core_validate_project(...)` - both bare, no `catch_unwind`/`spawn_blocking`.
- Contrast `crates/agnix-lsp/src/backend.rs` which handles the `JoinError` from `spawn_blocking` and emits an internal-error diagnostic instead of crashing.
- `Cargo.toml` release profile has no `panic = "abort"`, so the unwind propagates and aborts the connection task.

## Suggested fix

Wrap the MCP core-validation calls in `catch_unwind` (or run them via `spawn_blocking` and handle the `JoinError`) and convert a panic into an error result, mirroring the LSP backend, so a single panicking validator degrades one request instead of killing the server.

## Environment
- agnix version: 0.37.2

### LINKED ISSUE #1157: [BUG] VS Code binary downloader follows redirects with no host allowlist (JetBrains and Zed enforce one)
## Description

Of the three IDE-extension binary downloaders, only JetBrains and Zed validate that the initial URL and every followed redirect stay within a GitHub-owned host allowlist before writing/trusting the downloaded bytes. VS Code's `download-file.ts` follows up to `MAX_REDIRECTS = 10` via `res.headers.location` with no host check of any kind - it will transparently follow a redirect to any host.

The sha256 checksum verification in `extension.ts` is fetched via this same unchecked path (both the archive and its `.sha256` sidecar), so an attacker who can inject a redirect controls (or can co-host) both, making the checksum confirm tampering rather than prevent it. This is a defense-in-depth + cross-editor-consistency gap: the checksum is the current backstop, but the host-allowlist layer the other two editors have is absent here.

## Evidence

- `editors/vscode/src/download-file.ts`: `downloadFile()` follows `res.headers.location` up to `MAX_REDIRECTS = 10` with zero hostname/origin checks (grep for github.com/githubusercontent/trusted/allowlist/host returns no matches in the file).
- `editors/jetbrains/src/main/kotlin/io/agnix/jetbrains/binary/AgnixBinaryDownloader.kt:35` defines `TRUSTED_DOWNLOAD_HOSTS` and calls `isTrustedDownloadUrl()` on every hop (:44, :62).
- `editors/zed/src/lib.rs`: `fn is_trusted_download_url` gated before each `zed::download_file()`, backed by `trusted_github_url_accepted`/`untrusted_url_rejected`/`http_url_rejected` unit tests.

## Suggested fix

Port the JetBrains/Zed `TRUSTED_DOWNLOAD_HOSTS` check into `download-file.ts`: validate the initial URL and every redirect `location` against the GitHub host allowlist (`github.com`, `*.githubusercontent.com`) before following, rejecting any off-allowlist hop.

## Environment
- agnix version: 0.37.2

### LINKED ISSUE #1158: [BUG] Unmaintained serde_yaml 0.9.34+deprecated ships as a runtime transitive dependency, untracked by the three advisory mechanisms
## Description

The workspace deliberately routes its own YAML usage through the maintained `serde_norway` fork (`Cargo.toml:44` `serde_yaml = { package = "serde_norway", version = "0.9" }`), but a production dependency independently pulls the real, upstream-deprecated `serde_yaml 0.9.34+deprecated` into the runtime dependency graph via `rust-i18n`. Two parallel YAML stacks (`unsafe-libyaml` vs `unsafe-libyaml-norway`) are present in the lockfile.

This is a supply-chain hygiene gap: the deprecated crate is invisible to all three of the repo's tracking mechanisms (`deny.toml` ignore list, `security.yml`, `docs/RUSTSEC-ADVISORIES.md`) because no RUSTSEC advisory ID currently targets `serde_yaml`'s unmaintained status, so cargo-audit/cargo-deny report clean.

Note on risk (verified): the `serde_yaml` edge is a *runtime* (`-e normal`) transitive of `rust-i18n` (`rust-i18n -> rust-i18n-support -> serde_yaml`), so it links into the binary - but agnix's locales are compile-time-embedded and trusted, so this is a hygiene/footprint concern, not an untrusted-input deserialization risk.

## Evidence

- `cargo tree -i serde_yaml`: `serde_yaml v0.9.34+deprecated <- rust-i18n-macro/rust-i18n-support <- rust-i18n v4.1.0 <- agnix-core v0.37.2`.
- `cargo tree -e normal -p rust-i18n` shows `rust-i18n -> rust-i18n-support -> serde_yaml` as a runtime edge (not only the proc-macro).
- `crates/agnix-core/Cargo.toml:41` `rust-i18n = { workspace = true }` (production dep; `rust_i18n::i18n!` used in `lib.rs`).
- `deny.toml` ignore list contains only RUSTSEC-2025-0141 and RUSTSEC-2026-0173; `serde_yaml` is absent from all three trackers.

## Suggested fix

Either (a) upgrade/replace `rust-i18n` with a version that no longer pulls `serde_yaml`, or (b) if it must stay, document the deprecated transitive in `docs/RUSTSEC-ADVISORIES.md` and consider a `deny.toml` `[bans]` note so the coexistence is tracked rather than silent.

## Environment
- agnix version: 0.37.2

avifenesh:
## Summary
- harden MCP validation so core panics become MCP errors and diagnostic paths are workspace-relative
- extend panic isolation to project-level checks
- bind shell checksum parsing to the selected artifact filename and require trusted VS Code download redirect hosts
- correct the frontmatter parser safety comment and track the remaining deprecated transitive serde_yaml runtime dependency
- update the changelog for release notes

Closes #1149
Closes #1150
Closes #1154
Closes #1155
Closes #1156
Closes #1157
Closes #1158

## Validation
- cargo fmt --check --all
- cargo clippy --workspace --all-targets --all-features -- -D warnings
- cargo test --workspace --all-features
- cargo audit --deny warnings --ignore RUSTSEC-2025-0141 --ignore RUSTSEC-2026-0173
- cargo deny check advisories
- npm --prefix editors/vscode run test:unit
- node scripts/sync-rule-bookkeeping.js --check
- diff -q CLAUDE.md AGENTS.md
- git diff --check

### DIFF
--- a/crates/agnix-core/src/pipeline.rs
+++ b/crates/agnix-core/src/pipeline.rs
@@ -852,17 +852,17 @@ pub fn validate_project_with_registry(
     }
 
     // Run project-level checks (AGM-006, XP-004/005/006, VER-001)
-    {
+    diagnostics.extend(run_project_level_checks_with_panic_guard(&root_dir, || {
         agents_md_paths.sort();
         instruction_file_paths.sort();
 
-        diagnostics.extend(run_project_level_checks(
+        run_project_level_checks(
             &agents_md_paths,
             &instruction_file_paths,
             &config,
             &root_dir,
-        ));
-    }
+        )
+    }));
 
     sort_diagnostics(&mut diagnostics);
 
@@ -890,6 +890,28 @@ fn sort_diagnostics(diagnostics: &mut [Diagnostic]) {
             .then_with(|| a.suggestion.cmp(&b.suggestion))
             .then_with(|| a.fixes.len().cmp(&b.fixes.len()))
     });
+}
+
+fn run_project_level_checks_with_panic_guard<F>(root_dir: &Path, checks: F) -> Vec<Diagnostic>
+where
+    F: FnOnce() -> Vec<Diagnostic>,
+{
+    match std::panic::catch_unwind(std::panic::AssertUnwindSafe(checks)) {
+        Ok(diagnostics) => diagnostics,
+        Err(err) => {
+            let panic_message = panic_payload_message(err.as_ref());
+            vec![
+                Diagnostic::error(
+                    root_dir.to_path_buf(),
+                    0,
+                    0,
+                    "project::panic",
+                    t!("rules.file_panic_error", error = panic_message),
+                )
+                .with_suggestion(t!("rules.file_panic_error_suggestion")),
+            ]
+        }
+    }
 }
 
 #[cfg(feature = "filesystem")]
@@ -1407,6 +1429,24 @@ mod tests {
             }),
             "expected file::panic diagnostic, got {:?}",
             result.diagnostics
+        );
+    }
+
+    #[test]
+    fn project_level_panic_becomes_diagnostic() {
+        let temp = tempfile::TempDir::new().unwrap();
+
+        let diagnostics = run_project_level_checks_with_panic_guard(temp.path(), || {
+            panic!("intentional project panic")
+        });
+
+        assert_eq!(diagnostics.len(), 1);
+        assert_eq!(diagnostics[0].rule, "project::panic");
+        assert_eq!(diagnostics[0].file, temp.path());
+        assert!(
+            diagnostics[0].message.contains("intentional project panic"),
+            "expected panic payload in diagnostic: {:?}",
+            diagnostics[0]
         );
     }
 }
--- a/crates/agnix-core/src/pipeline.rs
+++ b/crates/agnix-core/src/pipeline.rs
@@ -625,12 +625,14 @@ pub fn validate_project_rules(root: &Path, config: &LintConfig) -> LintResult<Ve
     instruction_file_paths.sort();
 
     let mut diagnostics = files_config_diags;
-    diagnostics.extend(run_project_level_checks(
-        &agents_md_paths,
-        &instruction_file_paths,
-        &config,
-        &root_dir,
-    ));
+    diagnostics.extend(run_project_level_checks_with_panic_guard(&root_dir, || {
+        run_project_level_checks(
+            &agents_md_paths,
+            &instruction_file_paths,
+            &config,
+            &root_dir,
+        )
+    }));
     Ok(diagnostics)
 }
 
--- a/crates/agnix-core/src/pipeline.rs
+++ b/crates/agnix-core/src/pipeline.rs
@@ -878,6 +878,7 @@ pub fn validate_project_with_registry(
         .with_validator_factories_registered(validator_factories_registered))
 }
 
+#[cfg(feature = "filesystem")]
 fn sort_diagnostics(diagnostics: &mut [Diagnostic]) {
     diagnostics.sort_by(|a, b| {
         a.level
@@ -892,6 +893,7 @@ fn sort_diagnostics(diagnostics: &mut [Diagnostic]) {
     });
 }
 
+#[cfg(feature = "filesystem")]
 fn run_project_level_checks_with_panic_guard<F>(root_dir: &Path, checks: F) -> Vec<Diagnostic>
 where
     F: FnOnce() -> Vec<Diagnostic>,
--- a/scripts/download.sh
+++ b/scripts/download.sh
@@ -152,7 +152,28 @@ if [ "${HTTP_CODE}" != "200" ]; then
     exit 1
 fi
 
-EXPECTED_SHA="$(awk 'NF {print $1; exit}' "${CHECKSUM_FILE}" | tr '[:upper:]' '[:lower:]')"
+EXPECTED_SHA="$(awk -v expected="${ARTIFACT_NAME}" '
+NF {
+    hash = tolower($1)
+    file = $2
+    sub(/^\*/, "", file)
+    n = split(file, parts, /[\\\/]/)
+    base = parts[n]
+    if (base == expected) {
+        print hash
+        found = 1
+        exit
+    }
+}
+END {
+    if (!found) {
+        exit 1
+    }
+}
+' "${CHECKSUM_FILE}")" || {
+    echo "Error: Checksum file does not contain an entry for ${ARTIFACT_NAME}" >&2
+    exit 1
+}
 if ! printf '%s\n' "${EXPECTED_SHA}" | grep -Eq '^[0-9a-f]{64}$'; then
     echo "Error: Invalid checksum file for ${ARTIFACT_NAME}" >&2
     exit 1
--- a/scripts/download.sh
+++ b/scripts/download.sh
@@ -157,6 +157,7 @@ NF {
     hash = tolower($1)
     file = $2
     sub(/^\*/, "", file)
+    sub(/\r$/, "", file)
     n = split(file, parts, /[\\\/]/)
     base = parts[n]
     if (base == expected) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- **Security follow-ups** (closes #1149, #1150, #1154, #1155, #1156, #1157, #1158). Hardened MCP validation against handler panics and absolute-path disclosure, extended panic isolation to project-level checks, bound shell checksum parsing to the selected artifact filename, added VS Code release-download redirect host validation, corrected the YAML parser safety comment, and documented the remaining deprecated transitive `serde_yaml` dependency from `rust-i18n`.
+
 ## [0.37.2] - 2026-07-04
 
 ### Fixed
--- a/crates/agnix-core/src/parsers/frontmatter.rs
+++ b/crates/agnix-core/src/parsers/frontmatter.rs
@@ -9,9 +9,9 @@
 //!    extremely large YAML payloads from being read.
 //!
 //! 2. **Parser Library**: agnix's direct `serde_yaml` crate name is backed by
-//!    the maintained `serde_norway` package, which preserves the YAML parser API
-//!    and internal protections against excessive memory usage and stack overflow
-//!    from deeply nested structures.
+//!    the maintained `serde_norway` package, which preserves the YAML parser API.
+//!    The parser is not treated as the nesting or expansion-bomb boundary; agnix's
+//!    own size cap and depth guard below provide that protection.
 //!
 //! 3. **Memory Limit**: The entire file is bounded at 1 MiB, limiting total
 //!    memory consumption regardless of structure complexity.
--- a/crates/agnix-mcp/src/main.rs
+++ b/crates/agnix-mcp/src/main.rs
@@ -32,7 +32,10 @@ use rmcp::{
 };
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
+use std::any::Any;
 use std::collections::HashSet;
+use std::fmt::Display;
+use std::panic::{AssertUnwindSafe, catch_unwind};
 use std::path::{Path, PathBuf};
 
 const TOOL_ALIASES: &[(&str, &str)] =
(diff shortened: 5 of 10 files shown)


crates/agnix-core/src/pipeline.rs
```diff
@@ -892,6 +892,28 @@ fn sort_diagnostics(diagnostics: &mut [Diagnostic]) {
...
 }
 
+fn run_project_level_checks_with_panic_guard<F>(root_dir: &Path, checks: F) -> Vec<Diagnostic>
+where
+    F: FnOnce() -> Vec<Diagnostic>,
+{
+    match std::panic::catch_unwind(std::panic::AssertUnwindSafe(checks)) {
+        Ok(diagnostics) => diagnostics,
+        Err(err) => {
+            let panic_message = panic_payload_message(err.as_ref());
```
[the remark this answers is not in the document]
avifenesh: Thanks for checking this. This was fixed in the latest push: `panic_payload_message` is now defined in `crates/agnix-core/src/pipeline.rs`, the regression test asserts the original panic text, and the latest CI has compiled `agnix-core` successfully (`CI / test` and `CI / agnix` are green on commit `68f03e9`).

crates/agnix-core/src/pipeline.rs
```diff
@@ -852,17 +852,17 @@ pub fn validate_project_with_registry(
     }
 
     // Run project-level checks (AGM-006, XP-004/005/006, VER-001)
-    {
+    diagnostics.extend(run_project_level_checks_with_panic_guard(&root_dir, || {
```
avifenesh, line 857: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

This wraps the full-project path, but the other project-level entry point is still unguarded: `validate_project_rules()` in this same file still calls `run_project_level_checks(...)` directly, and the LSP uses that API from `crates/agnix-lsp/src/backend/revalidation.rs`. A panic in an AGM/XP/VER project-level check can therefore still unwind through LSP project validation instead of becoming the new `project::panic` diagnostic. Please route `validate_project_rules()` through this guard as well (or share a guarded helper) so panic isolation covers all project-level checks.

scripts/download.sh
```diff
@@ -152,7 +152,28 @@ if [ "${HTTP_CODE}" != "200" ]; then
     exit 1
 fi
 
-EXPECTED_SHA="$(awk 'NF {print $1; exit}' "${CHECKSUM_FILE}" | tr '[:upper:]' '[:lower:]')"
+EXPECTED_SHA="$(awk -v expected="${ARTIFACT_NAME}" '
+NF {
+    hash = tolower($1)
+    file = $2
```
avifenesh, line 158: *This is an auto review done by [revuto](https://github.com/avifenesh/revuto).*

---

The filename match does not tolerate CRLF checksum sidecars. Windows release checksum files are generated with PowerShell `Out-File` in `.github/workflows/release.yml`, so the entry can end in `agnix-...zip\r`; `awk`'s default field splitting does not strip that carriage return, leaving `base` as `...zip\r` and making `base == expected` fail even for the correct artifact. Strip a trailing `\r` from `file`/`base` before comparing so Windows users do not get a false “Checksum file does not contain an entry” error.

avifenesh (commented):

---

Found two correctness issues in the hardening changes: one panic-isolation path is still uncovered, and the checksum filename matching can reject valid Windows sidecars.

crates/agnix-core/src/pipeline.rs
avifenesh: Thanks, good catch. Addressed in 307e65b by routing validate_project_rules() through run_project_level_checks_with_panic_guard(), so the LSP path gets project::panic diagnostics too. Re-ran cargo clippy --workspace --all-targets --all-features -- -D warnings and cargo test --workspace --all-features.

avifenesh:
Thanks for the review. The valid follow-ups have been addressed in the latest commit: project-level panic isolation now routes through the shared guard, and download.sh strips trailing CR from checksum sidecar filenames before comparison. The related review threads are resolved, and I re-ran cargo fmt --check --all, the targeted tests, clippy, the full workspace test suite, sync-rule-bookkeeping --check, AGENTS/CLAUDE parity, and git diff --check.

PR #1177, opened 2026-07-04
docs: version 0.37.3 docs

avifenesh:
Auto-generated versioned docs snapshot for v0.37.3, including updated rule docs and site data.


avifenesh:
Thanks, this is the expected generated docs snapshot for v0.37.3. Checks are green.

PR #1290, opened 2026-07-31
fix(ci): make the Claude review job actually review [skip changelog]

avifenesh:
The `claude-review` job has reported success in ~11 seconds and posted **nothing** on every PR for months — including all six drift PRs merged today. CLAUDE.md rule 11 treats it as "the major quality gate, and most thorough review", so this was a green check gating nothing.

Two separate defects.

## 1. No `prompt` on a `pull_request` event, so there was nothing to do

The job triggers on `pull_request`. With no `prompt` input, the action looks for an `@claude` trigger, finds none, has nothing to do, and exits successfully. That is the entire 11 seconds.

The fix is `track_progress: true`, which the action documents as:

> Force tag mode with tracking comments for **pull_request** and issue events.

That's the same mechanism the action's own [`examples/pr-review-comprehensive.yml`](https://github.com/anthropics/claude-code-action/blob/main/examples/pr-review-comprehensive.yml) uses for exactly this shape of job.

Added a prompt scoped to this repo rather than a generic one, in priority order:

1. **Correctness** — for a validator change, does the rule still reject what it should while accepting what the upstream doc endorses?
2. **Evidence** — every rule claim must trace to a cited source; flag behavior no `evidence.source_urls` entry supports, and any autofix that could change a config's *meaning* rather than its spelling.
3. **Test quality** — tests must assert both directions; a happy-path-only test is a gap.
4. **Bookkeeping** — `rules.json` ↔ `VALIDATION-RULES.md`, rule counts, locales, `AGENTS.md`/`CLAUDE.md` parity.

Formatting nits are explicitly excluded, since `cargo fmt` and clippy already gate them — otherwise the review spends its attention on noise.

The three items after correctness are the failure classes that actually showed up in today's work: two rules had no upstream origin at all, and one autofix silently changed when a hook fired.

## 2. `model` is not an action input

Both workflows passed `model: claude-opus-4-6` as an input. The logs show:

```
##[warning]Unexpected input(s) 'model', valid inputs are ['trigger_phrase', ...]
```

It was ignored, so any run that *had* worked would have used the default model, not the pinned one. It belongs in `claude_args` as `--model`.

## `claude.yml` deliberately keeps no prompt

That workflow fires on an explicit `@claude` mention, so the comment body *is* the instruction — adding a prompt would override what the user asked for. It only needed the `model` fix.

## Verification

- Both files parse as YAML.
- Cross-checked every remaining input against the action's own `action.yml` input list: **no unknown keys** in either workflow. That check is what would have caught the original bug.
- The real proof is behavioral and arrives on this PR: `claude-review` should now take minutes rather than 11 seconds and leave an actual review. I'll confirm before merging.

`[skip changelog]`: CI-only, no user-facing behavior change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,6 +69,14 @@ jobs:
         # from knowledge-base/rules.json.
         run: python3 scripts/check-rule-counts.py
 
+      - name: Locale sync check
+        if: matrix.os == 'ubuntu-latest'
+        # The i18n!() macro loads from each crate's own locales/ directory, so a
+        # diagnostic string edited only in the root copy silently never reaches
+        # the binary. Nothing gated this: the script existed but no workflow ran
+        # it.
+        run: bash scripts/check-locale-sync.sh
+
       - name: Schema sync check
         if: matrix.os == 'ubuntu-latest'
         # Fails if the committed .agnix.toml schema differs from the current
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,9 @@ jobs:
 
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2
         with:
+          # `shared-key` (rather than the default job-name key) lets the Claude
+          # review job restore this cache instead of cold-building the workspace.
+          shared-key: agnix-workspace
           save-if: ${{ github.ref == 'refs/heads/main' }}
 
       - name: Format check
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -103,10 +103,17 @@ jobs:
           components: clippy, rustfmt
 
       # Restore ci.yml's cache via the shared key. Without it the first
-      # `cargo test` cold-builds a 6-crate workspace, which is slow enough that
-      # the reviewer would skip running tests and fall back to reading them -
-      # putting prompt priorities 1 and 3 back where they started. Read-only:
-      # this job should consume ci.yml's cache, never publish its own.
+      # `cargo test` cold-builds the third-party dependency graph, which is slow
+      # enough that the reviewer would skip running tests and fall back to
+      # reading them - putting prompt priorities 1 and 3 back where they started.
+      #
+      # Scope: rust-cache cleans workspace-member artifacts before saving, so it
+      # caches dependencies only. A warm restore still leaves the 6 workspace
+      # crates and every test binary to compile per review - budget that against
+      # `timeout-minutes` above. The cache is also empty until this merges and
+      # one `main` run populates the new key, and adding `shared-key` abandons
+      # ci.yml's old job-name-keyed entry, so CI itself cold-builds for one merge
+      # cycle. Read-only here: this job consumes ci.yml's cache, never publishes.
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2
         if: steps.check-runs.outputs.should_run == 'true'
         with:
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -92,17 +92,26 @@ jobs:
           # Passing it as an input produced `Unexpected input(s) 'model'` and was
           # silently ignored, so reviews would have run on the default model.
           # The pinned value was also a stale model ID.
-          # No `--allowedTools`: an explicit list here blocked the checks the
-          # prompt below asks for. Verifying that a rule's `evidence.source_urls`
-          # actually supports its behavior needs WebFetch, and judging validator
-          # correctness on a Rust workspace needs `cargo test`; an allowlist
-          # without them reduces both to reading rather than checking. It also
-          # omitted `mcp__github_comment__update_claude_comment`, the tool the job
-          # posts through, which only worked because the action injects it anyway.
-          # The action's tag-mode defaults already grant the GitHub MCP tools plus
-          # Read/Grep/Glob.
+          # `--allowedTools` is ADDITIVE, not restrictive: per the action's
+          # configuration docs, "The base GitHub tools are always included. Use
+          # `--allowedTools` to add additional tools (including specific Bash
+          # commands)". So omitting it grants nothing extra - it only drops what
+          # the list itself would have added, and the baseline still withholds
+          # WebFetch and cargo.
+          #
+          # That matters because the prompt below asks for checks the baseline
+          # cannot perform:
+          #   - WebFetch/WebSearch: priority 2 is verifying that a rule's
+          #     `evidence.source_urls` actually says what the rule claims. Without
+          #     these, a reviewer can only confirm a URL is present.
+          #   - cargo test/clippy: priorities 1 and 3 are validator correctness
+          #     and whether tests assert both directions, which on a Rust
+          #     workspace means running them, not reading them.
+          #   - the inline-comment tool: the prompt asks for inline comments, and
+          #     the baseline provides only the top-level comment tool.
           claude_args: |
             --model claude-opus-5
+            --allowedTools "WebFetch,WebSearch,Bash(cargo test:*),Bash(cargo clippy:*),Bash(node scripts/sync-rule-bookkeeping.js:*),mcp__github_inline_comment__create_inline_comment"
           prompt: |
             REPO: ${{ github.repository }}
             PR NUMBER: ${{ github.event.pull_request.number }}
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -36,25 +36,19 @@ jobs:
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           fetch-depth: 0
-          # This job now runs `cargo test` on PR-authored code. Match ci.yml and
-          # do not leave a usable token in the checkout's git config.
+          # Kept, but it does NOT leave this job token-free. Dropping
+          # `use_commit_signing` puts the action in non-signing mode, and that
+          # branch calls `configureGitAuth` (src/modes/tag/index.ts:88), which
+          # unsets checkout's extraheader and then embeds the token in the remote
+          # URL (src/github/operations/git-config.ts:78). Verified in a live run:
+          # `.git/config` carries an `x-access-token` entry the action added after
+          # checkout. The token-free path there needs ALLOWED_NON_WRITE_USERS,
+          # which never engages behind the author_association gate above.
+          #
+          # So this only closes the window before the action runs. That is worth
+          # having and costs nothing, but the job still compiles PR-authored code
+          # alongside a readable token holding pull-requests/issues write.
           persist-credentials: false
-
-      - uses: dtolnay/rust-toolchain@4be9e76fd7c4901c61fb841f559994984270fce7 # stable
-        with:
-          # Pinned like ci.yml rather than relying on the runner image, since
-          # `cargo clippy` is in the grant list below.
-          components: clippy, rustfmt
-
-      # Restore ci.yml's cache via the shared key. Without it the first
-      # `cargo test` cold-builds a 6-crate workspace, which is slow enough that
-      # the reviewer would skip running tests and fall back to reading them -
-      # putting prompt priorities 1 and 3 back where they started. Read-only:
-      # this job should consume ci.yml's cache, never publish its own.
-      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2
-        with:
-          shared-key: agnix-workspace
-          save-if: "false"
 
       - name: Check run count
         id: check-runs
@@ -98,6 +92,26 @@ jobs:
             echo "should_run=false" >> $GITHUB_OUTPUT
             echo "[SKIP] Skipping Claude review (already ran 3 times for this commit)"
           fi
+
+      # Toolchain and cache come after the run-count gate: both are pointless
+      # work when the review is going to be skipped.
+      - uses: dtolnay/rust-toolchain@4be9e76fd7c4901c61fb841f559994984270fce7 # stable
+        if: steps.check-runs.outputs.should_run == 'true'
+        with:
+          # Pinned like ci.yml rather than relying on the runner image, since
+          # `cargo clippy` is in the grant list below.
+          components: clippy, rustfmt
+
+      # Restore ci.yml's cache via the shared key. Without it the first
+      # `cargo test` cold-builds a 6-crate workspace, which is slow enough that
+      # the reviewer would skip running tests and fall back to reading them -
+      # putting prompt priorities 1 and 3 back where they started. Read-only:
+      # this job should consume ci.yml's cache, never publish its own.
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2
+        if: steps.check-runs.outputs.should_run == 'true'
+        with:
+          shared-key: agnix-workspace
+          save-if: "false"
 
       - name: Run Claude Code Review
         if: steps.check-runs.outputs.should_run == 'true'
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -41,6 +41,10 @@ jobs:
           persist-credentials: false
 
       - uses: dtolnay/rust-toolchain@4be9e76fd7c4901c61fb841f559994984270fce7 # stable
+        with:
+          # Pinned like ci.yml rather than relying on the runner image, since
+          # `cargo clippy` is in the grant list below.
+          components: clippy, rustfmt
 
       # Restore ci.yml's cache via the shared key. Without it the first
       # `cargo test` cold-builds a 6-crate workspace, which is slow enough that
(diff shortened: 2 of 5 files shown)


avifenesh:
All five findings addressed in 64ebeae2. The job reviewing its own fix - and finding real problems in it - is the clearest confirmation it works.

**1-3, `--allowedTools`: dropped entirely.** Your tool-denial table settled it. An allowlist without `WebFetch` reduces the evidence check to confirming a URL is *present*, never that it says what the rule claims — the one check this repo most needs a reviewer for. Same for `cargo test` and correctness. You were also right that it omitted `mcp__github_comment__update_claude_comment` (working only because the action injects it) and that `gh pr comment` hands the model a way to do the opposite of "one top-level comment". Taking the tag-mode defaults.

**4, concurrency + wording:** added the group with `cancel-in-progress`, matching `docs-site.yml`. And you were right that the throttle is per-commit, not per-PR — the `head_sha` filter resets it every push. Corrected both the log line and the user-facing skip comment, and documented that the `status=completed` filter makes in-flight runs invisible, so the concurrency group is what actually prevents overlap.

**5, `use_commit_signing`:** dropped, and narrowed the job to `contents: read` since it no longer commits. A review-only job shouldn't carry `commit_files`. `claude.yml` keeps it — that workflow is meant to make changes when asked.

One correction to the PR body's framing: `--allowedTools` *was* a third undescribed change, as you noted. That's on me — it went in as boilerplate copied from the action's example without checking it against the prompt shipped beside it.

avifenesh:
You were right and I was wrong. Fixed in c25f08de.

I verified your claim against the action's own configuration docs before reverting, and it's stated outright:

> **Note**: The base GitHub tools are always included. Use `--allowedTools` to add additional tools (including specific Bash commands), and `--disallowedTools` to prevent specific tools from being used.

Additive. Your empirical table was the giveaway I should have caught myself — `node` and `git` were allowed in the *first* run despite never being in the list, which only makes sense if the list adds rather than replaces. Removing it granted nothing and cost the inline-comment tool.

Restored with what the prompt actually needs: `WebFetch`, `WebSearch`, `Bash(cargo test:*)`, `Bash(cargo clippy:*)`, `Bash(node scripts/sync-rule-bookkeeping.js:*)`, and `mcp__github_inline_comment__create_inline_comment` — the last one closing finding 2, so the "use inline comments" instruction is fulfillable again.

Took your suggestion to record the additive semantics in the comment block rather than just fixing the line. The misreading was the durable hazard, not the missing tools.

Worth noting what happened here: your first review's finding 1-3 diagnosis was wrong, you re-tested it against the running config, and corrected yourself. That's a better outcome than either review alone, and it's the behavior that makes this gate worth having.

avifenesh:
All three addressed in 52903c62, plus both items you raised as notes.

**1, `claude.yml` can write but not test.** The strongest of the three — same additive bug this PR exists to fix, sitting one file over, and worse there because that workflow actually commits. Granted cargo test/clippy/fmt/check/build/run plus the three bookkeeping scripts. An `@claude fix X and add a test` request will now compile what it pushes.

**2, the cargo grant was permitted but not usable.** You were right that a cold 6-crate build means the reviewer rationally skips testing and goes back to reading — restoring the exact state the grant was meant to end. Added the stable toolchain and pointed rust-cache at `ci.yml`'s cache. Your detail about rust-cache keying on job name was the load-bearing part: I added `shared-key: agnix-workspace` to **both** sides, since setting it here alone would have silently missed. `save-if: "false"` so this job consumes that cache rather than publishing a competing one.

**3, cancelled runs burning budget.** Filtering on `conclusion == "success"`. Worth noting this was a bug *I* introduced last round — the concurrency group created it — and your correction of your own earlier "different head_sha" claim is what made it visible. Draft → ready_for_review on the same sha was the case I'd have missed.

**Both notes taken rather than deferred:** `persist-credentials: false` (this job now executes PR-authored code via cargo, so matching `ci.yml` is right), and `timeout-minutes: 30` so a hung run can't hold the concurrency group.

Three rounds, and each one found something real that the previous round introduced or missed. For a check that produced nothing at all this morning, that's the outcome worth having.

avifenesh:
Both addressed in 81aa2020, and your third note turned out to be a real gap worth closing.

**1, `claude.yml` grant without toolchain or cache.** Caught cleanly — I fixed "permitted but not usable" in one file while creating it in the other, in the same commit. Added both steps plus `timeout-minutes: 60`, since an `@claude` request can now trigger a cold build against a 6-hour default.

**2, missing bookkeeping gates.** Granted `check-rule-counts.py`, `check-locale-sync.sh`, `cargo check`, and `cargo build`. You were right that `check-rule-counts.py` covers drift `sync-rule-bookkeeping.js` doesn't — it's a separate `ci.yml` gate, and denying it to a reviewer whose priority 4 is rule-count parity was incoherent.

**3, `components: clippy, rustfmt`** pinned on both toolchain steps rather than relying on the runner image.

**On your note that `check-locale-sync.sh` runs in no workflow** — I verified that and it's a real hole, so I've gated it in `ci.yml` rather than just granting it. `rust_i18n` loads from each crate's own `locales/` directory, so a diagnostic string edited only in the root copy never reaches the binary. That exact thing happened earlier today (#1283) and was caught by hand when a script failed for an unrelated reason. Now it can't reach main silently.

**On the empty cache:** agreed, and worth being explicit — it stays cold until this merges and one `main` run populates the key. So `cargo test` is genuinely expensive for the next review or two, including possibly the one on this comment. Self-resolving, not something to design around.

Four rounds, each finding something the previous one introduced or missed — including two bugs of mine created *by* the earlier fixes. That's the gate doing its job.

avifenesh:
Both addressed in 8743d2db.

**1, `cargo run`.** You were right that this isn't a nice-to-have — the eval suite is the one gate `cargo test` doesn't cover, and "a validator change that rescopes which rules fire" describes most of what this repo does. Granted. `cargo fmt` stays out for the reason you gave.

**2, the persisted credential.** You flagged this as check-don't-change-blind, which was the right call, so I checked rather than assuming. With `use_commit_signing: true` the action commits through GitHub's API — its own docs describe that mode as "simple but cannot perform complex git operations like rebasing", which is only true of an API-based path. The credential is unused, so dropping it is free. Taken.

That closes everything from five rounds. Merging once checks are green.

Worth stating what this thread demonstrates, since it's the whole point of the PR: the job caught two bugs I introduced *with the fixes themselves* (the concurrency group leaking the run budget, and granting cargo to `claude.yml` without making it usable), corrected its own wrong diagnosis about `--allowedTools` after re-testing, and turned up a real hole nobody was looking for — `check-locale-sync.sh` existing but running in no workflow. That last one had already bitten today and was caught by hand.

This morning the same check passed in 11 seconds and said nothing.

avifenesh:
Both addressed in 7156a657. The first one was a comment of mine stating something false, which is the worst kind to leave on a security control.

**1, the credential claim.** I verified your source trace rather than taking it:

- `src/modes/tag/index.ts:71` — `useApiCommitSigning = useCommitSigning && !useSshSigning` → false once `use_commit_signing` is dropped
- `:88` — `} else if (!useApiCommitSigning) { await configureGitAuth(...) }`
- `src/github/operations/git-config.ts:47` — unsets checkout's extraheader
- `:77-79` — `else` branch builds `https://x-access-token:${githubToken}@...` and runs `git remote set-url origin`

The token-free path above it needs `ALLOWED_NON_WRITE_USERS`, which never engages behind the `author_association` gate. So the line closes only the pre-action window. Kept it — free, and that window is real — but the comment now says that instead of claiming a property the job doesn't have.

Your inversion point was the sharpest part: the claim *does* hold in `claude.yml`, and for a reason the docs quote didn't give — `use_commit_signing: true` means `useApiCommitSigning` is true, so `configureGitAuth` is never reached. I rewrote that comment to cite the mechanism and cross-reference the review job, so the two `persist-credentials: false` lines aren't read as inconsistent when they behave differently.

**2, step ordering.** Correct and embarrassing — toolchain and cache ran before the gate deciding whether to review at all, so a skipped run still paid for both. Now gated on `should_run`.

Six rounds. This one found a false security comment I wrote and wasted setup work I introduced two commits earlier. Merging once green.

avifenesh:
All three addressed in daeaadb4. No config defects this round — all three were precision problems in prose I wrote, which is the right thing to be left with.

**1, `claude.yml`'s credential comment.** Same class as the false claim you caught last round, one file over: it opened with "this workflow now compiles PR-authored code" and then described a git-config mitigation, reading as complete. Your point about `DEFAULT_WORKFLOW_TOKEN`/`OVERRIDE_GITHUB_TOKEN` in the action's `action.yml` is right, and I agree with not changing config over it — removing the cargo grant recreates "commits code it never compiled". The comment now states its scope and names the `author_association` gate as what actually bounds the risk.

I'll also note you were explicit that the env-inheritance step was source-verified plus inference, not measured, and that you declined to write an environment-dumping probe on a live job to close the gap. Flagging the limit of your own evidence is more useful than a confident claim would have been.

**2, `docs/TRANSLATING.md`.** You re-raised this after I missed it, and it was worse than a stale name: false before this PR, then true-in-substance-wrong-in-name after. Both mentions now say "Locale sync check" step in the `ci` job, so grep finds it.

**3, the cache comment.** Right that rust-cache cleans workspace members and caches dependencies only, so the workspace crates and test binaries still compile per review. Recorded that, plus the two transitional facts: empty until this merges, and `shared-key` abandons `ci.yml`'s old job-name entry so CI cold-builds for one cycle.

Merging once green. Seven rounds on a check that this morning passed in 11 seconds having done nothing — and across them it caught a false security comment, two bugs introduced by its own earlier fixes, a doc claim that was wrong before and after, and a CI gate that existed as a script but ran nowhere.

avifenesh:
All three notes closed in 512d4796, including the one you deliberately left as a decision.

**1, ungated setup in `claude.yml`.** You were right not to propose `github.event_name != 'issues'` — an issue-triggered `@claude` can create a branch and commit, so gating on event type fails silently in the one direction that matters. Decided to keep it ungated and recorded *why* in the comment, so the asymmetry with the review job reads as intentional rather than an oversight.

**2, `TRANSLATING.md`.** Right that `ci.yml:14` sets `name:` on the matrix, so `ci` never appears in the UI. Now says it shows as `test (ubuntu-latest)`. Also dropped the ` -- ` you spotted; the remaining one is `cargo test -- i18n_tests`, which rule 14 exempts as a CLI separator.

**3, the locale gate's own blind spot.** You said it wasn't this PR's job, but it's worth doing now that the script is a live gate rather than dead code — and your framing was exact: priority 3's "assert both directions" applied to the gate itself. A locale added to a crate but never to root *would ship*, since `rust_i18n` loads from the crate copy while the canonical root copy silently lacked it. Added the reverse pass and verified both ways: a crate-only `fr.yml` now fails with a pointed message, clean tree still passes.

Merging.

Eight rounds. This check passed in 11 seconds this morning having done nothing. Across the rounds it caught a false security comment I'd written, two bugs introduced by its own earlier fixes, a doc claim wrong both before and after the change, and a CI script that existed but ran nowhere — then corrected its own wrong diagnosis about `--allowedTools` after re-testing, and flagged the limits of its own evidence on the env-inheritance question instead of overclaiming.

avifenesh:
All three residuals closed in 580e79ac, plus the prose note.

**1, the remediation footer.** Correct and a good catch — the printed command copies outward, so for the crate-only case it's a no-op and a contributor would get identical output on re-run. Now tracks the reverse count separately and prints the inward copy first when it applies.

**2, `CRATES` derivation.** You retracted your own earlier one-line suggestion after finding it would drop the "directory missing" check — that's the more useful contribution than the original suggestion was. Went with floor-plus-discovery: the explicit three keep their failure mode for a deleted `locales/`, unioned with anything discovered so a future crate is covered without editing the list.

**3, `*.yml` only.** Right, and you verified it against the rust-i18n docs rather than asserting it. Both loops now cover `.yml`/`.yaml`/`.json`/`.toml`, so a deviation from the `.yml` convention fails loudly instead of being loaded by the macro while invisible to the gate.

**On your stated verification limit** — you said plainly that your grant couldn't create fixtures, so you'd verified the reverse branch by reading and that "the empirical half is the author's, not mine." Took that literally and exercised all five branches:

| case | result |
|---|---|
| crate-only `fr.yml` | fails, inward hint printed |
| crate-only `de.json` | fails (new extension coverage) |
| root `it.yml` missing from a crate | fails, outward hint printed |
| deleted `crates/agnix-lsp/locales/` | still fails (floor preserved) |
| clean tree | passes |

**Prose note:** split into two labelled points. You were right that the splice partly undid two rounds of wording on that block — worth fixing precisely because of how much care went into it.

Merging now.

avifenesh:
Addressed in 5030b3a1, plus both notes.

**1, the footer.** You were right twice over. It printed `.yml` copy commands for a `.json` failure — the same no-op-hint loop the previous commit set out to close, one extension over. And the advice was wrong in substance: the widened globs exist so a convention deviation *fails loudly*, then the footer told you to propagate the deviation. `docs/TRANSLATING.md` mandates `<code>.yml`, so it now advises a rename. The fan-out is derived from `CRATES` too, so a discovered crate can't be excluded from advice about a check that covered it.

Verified all four branches: `.json` → rename advice; crate-only `.yml` → inward copy plus derived fan-out; discovered 4th crate → appears in the printed command; clean → passes.

**Notes:** "sets the same flag" now names `persist-credentials` explicitly — you were right it read as self-contradicting next to "dropping". And `TRANSLATING.md` now points at the gate rather than relying on its hardcoded crate loop, since that list was the third copy and the one that could silently disagree.

**On your correction of your own stated limit** — you said last round your grant couldn't create fixtures, then found this round that it can, and said so before presenting the measured results. Retracting your own constraint claim, unprompted, is the same move as the `--allowedTools` retraction in round 3. That's the behavior that made ten rounds worth running rather than one.

Merging now.

avifenesh:
Fixed in e9539698 — and you were right that it's the exact mirror of what the previous commit fixed. I closed the rename path for the crate direction and left the root direction falling through to a copy hint that can't move a `.json`.

**1, root-direction hole.** Root-side deviations are now counted and reported, so `locales/zz.json` gets rename advice.

**Note on exclusive branches.** Also right, and the sharper of the two: the failure *classes* aren't exclusive but the branches were, so a mixed tree printed only the rename block with the fan-out described in prose and no command. Each applicable class now prints its own step, and the fan-out always prints since it's the last step for all of them.

**Third mention, now actually fixed:** the empty-directory message named `.yml` while the scan covers four extensions.

Verified all five branches with fixtures:

| case | printed |
|---|---|
| root `zz.json` | rename + fan-out |
| crate `aa.json` **and** crate `bb.yml` | rename + inward copy + fan-out |
| plain forward drift | fan-out only |
| empty `locales/` | names all four extensions |
| clean | passes |

Merging now — all checks green.

Eleven rounds on a check that this morning passed in 11 seconds having done nothing. Along the way it caught a false security comment I wrote, four bugs introduced by its own earlier fixes, a doc claim wrong both before and after, and a CI script that existed but ran nowhere — and twice retracted its own claims after re-testing (the `--allowedTools` diagnosis, and its stated inability to create fixtures). The last three rounds were entirely me fixing one class of bug in one direction at a time while it found the other.

avifenesh:
Fixed in 6d1f2b8a. Both, plus your verification limit.

**1, the per-file/footer contradiction.** Exactly right, and the sharpest framing yet: `expected copy of` is an instruction to create three more copies of the deviation the globs exist to catch. Line 42 appended the file before judging its extension. Now counted and skipped — the loud failure stays, the contradiction and the inflated `4 ... out of sync` go. One misnamed root file now reports exactly one error with rename advice and no copy paths.

**Note, footer lead-in.** Right — dropping the old `else echo "Run:"` left plain drift opening on "Then ..." with nothing before it. Reworded conditionally.

**Your verification limit.** You said staging an empty `locales/` needs moving the real directory and your grant declines that. Closed it rather than leaving it as mine-to-do: built a sandbox tree in `/tmp` with the script and empty dirs, which exercises the branch without touching the repo. Result — `FAIL: No locale files (*.yml *.yaml *.json *.toml) found in locales/`, exit 1, no glob expansion. Your read was correct on both points.

Merging.

Twelve rounds. The last four were the same class of bug — advice that contradicted the check — surfacing one corner at a time: crate direction, then root direction, then the mixed case, then the per-file lines. Each time I fixed the instance and you found the next. That's a better outcome than any single pass, and it's worth noting the check produced literally nothing this morning.

avifenesh:
Fixed in 757e7a25. Reproduced your sandbox case first, then fixed it.

**1, the all-misnamed root.** Exactly as you measured: `en.json` matches `*.json`, the message named that glob while claiming none found, and `root_non_yml` had already counted it one loop earlier. The rename advice sat 80 lines past the exit. Now the two states are distinguished — all-misnamed reports the count it measured plus the rename command; a genuinely empty directory keeps the original message. Both verified, plus the real tree.

**Note, `TRANSLATING.md`.** Dropped the hardcoded "root + 3 crates", which sat two lines below the paragraph I'd added specifically to stop trusting a hardcoded list.

**On your limit correction:** you said last round that staging an empty `locales/` was outside your grant, then found you could build a sandbox inside the repo and did — and that's what surfaced this finding. Second time you've retracted a stated constraint with evidence attached rather than leaving it as an assumption. The limit being wrong is less interesting than the fact that checking it found a real bug.

Merging.

Thirteen rounds. The last five were one class — advice contradicting its own check — surfacing a corner at a time: crate direction, root direction, mixed case, per-file lines, all-misnamed. I fixed each instance; you found the next. Worth restating that this check passed in 11 seconds this morning and said nothing at all.

PR #1291, opened 2026-07-31
fix(ci): make the review cap per-PR and require exhaustive findings [skip changelog]

avifenesh:
Two problems with what #1290 shipped. Both were visible in that PR itself, which took **13 review rounds and 15 successful review runs** — under a cap that says 3.

## 1. The cap was per-commit, not per-PR

The API query filtered on `head_sha`, so the counter reset on every push. A PR could take unlimited reviews at three per commit, which is the opposite of what a cap is for.

Worse, when the reviewer flagged this in #1290 I **corrected the wording to match the bug** — changed the log line and skip comment to say "for this commit" and moved on. The count was the thing to fix, not the label.

Now scoped with `branch=<head ref>`, which spans the PR's whole life. Verified:

```
$ # against #1290's branch
paginated per-PR count = 15   (13 rounds + 2)
$ # unknown branch
count=[0]                     (not empty, so the numeric guard holds)
```

`--paginate` plus a summing `awk` because `--paginate` emits one JSON result per page — a PR past 100 runs would otherwise count only the last page. Given #1290 hit 15, that ceiling is not as far off as it sounds.

## 2. The prompt never asked for completeness

That is why 13 rounds happened. Five of them were **one class of bug** — advice contradicting its own check — surfacing a corner at a time:

| round | corner found |
|---|---|
| 9 | crate direction: `cp` outward can't fix a crate-only file |
| 11 | root direction: same hole, other side |
| 11 | mixed case: exclusive branches, non-exclusive failures |
| 12 | per-file lines: `expected copy of` told you to propagate the deviation |
| 13 | all-misnamed: reported "none found" for files it had just counted |

Each round found one instance, I fixed that instance, the next round found the next. Under a *real* per-PR cap, that pattern burns the entire budget without converging — which is exactly what fixing item 1 would have caused.

The prompt now requires sweeping the class before reporting, with the specific generalizations that would have collapsed those five rounds into one:

- a list missing an entry → check parallel lists for the same omission, and that list for other omissions
- a message contradicting its code → grep for other places making the same claim
- a check handling one direction → ask what the opposite direction does
- a fix in one file → check sibling files

It also now asks for confidence and verification method per finding, and to say plainly when something could not be verified rather than presenting inference as measurement. The reviewer did this unprompted in #1290 twice — flagging that env-inheritance was source-verified-plus-inference, and that a fixture was impractical — and both times it was more useful than a confident claim. Worth making explicit rather than hoping for it.

## Verification

- YAML parses; branch-scoped count verified against a real branch and an unknown one.
- The `conclusion == "success"` filter is retained — cancelled runs still report `status=completed`, so without it the concurrency group's cancellations would consume budget.
- No behavior change to the review itself beyond the prompt; `--allowedTools`, `--model`, `track_progress`, permissions, and the concurrency group are untouched.

`[skip changelog]`: CI-only.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -40,7 +40,7 @@ actions/setup-node@v6.2.0:      6044e13b5dc448c55e2357c09f80417699197238
 actions/configure-pages@v5:    983d7736d9b0ae728b81ab479565c72886d7745b
 actions/upload-pages-artifact@v4.0.0: 7b1f4a764d45c48632c6b24a0339c27f5614fb0b
 actions/deploy-pages@v4:       d6db90164ac5ed86f2b6aed7e0febac5b3c0c03e
-rhysd/actionlint@v1.7.1:       62dc61a45fc95efe8c800af7a557ab0b9165d63b
+rhysd/actionlint@v1.7.12:      914e7df21a07ef503a81201c76d2b11c789d3fca
 
 # Rust Tooling
 dtolnay/rust-toolchain@stable: 4be9e76fd7c4901c61fb841f559994984270fce7
@@ -63,7 +63,7 @@ softprops/action-gh-release@v2: a06a81a03ee405af7f2048a818ed3f03bbf83c7b
 huacnlee/zed-extension-action@v2: 8cd592a0d24e1e41157740f1a529aeabddc88a1b
 
 # Claude Code
-anthropics/claude-code-action@v1: 6867bb3ab0b2c0a10629b6823e457347e74ad6d2
+anthropics/claude-code-action@v1: e0cf66d1d257526b5d07f141838c338921cb8455
 ```
 
 ## Updating Action Versions
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -28,39 +28,39 @@ cache poisoning from pull requests.
 
 ## SHA Pin Reference
 
-When updating actions, use these SHA commits (last verified: 2026-07):
+When updating actions, use these SHA commits. Generated from the workflow
+files themselves - if this table and a `uses:` line disagree, the workflow
+is authoritative (last verified: 2026-07-31).
 
 ```yaml
 # GitHub Official Actions
-actions/checkout@v4:           34e114876b0b11c390a56381ad16ebd13914f8d5
-actions/upload-artifact@v4:    ea165f8d65b6e75b540449e92b4886f43607fa02
-actions/download-artifact@v7.0.0: 37930b1c2abaa49bbe596cd826c3c89aef350131
-actions/setup-python@v6.2.0:    a309ff8b426b58ec0e2a45f0f869d46889d02405
-actions/setup-node@v6.2.0:      6044e13b5dc448c55e2357c09f80417699197238
-actions/configure-pages@v5:    983d7736d9b0ae728b81ab479565c72886d7745b
-actions/upload-pages-artifact@v4.0.0: 7b1f4a764d45c48632c6b24a0339c27f5614fb0b
-actions/deploy-pages@v4:       d6db90164ac5ed86f2b6aed7e0febac5b3c0c03e
-rhysd/actionlint@v1.7.12:      914e7df21a07ef503a81201c76d2b11c789d3fca
+actions/add-to-project@v2.0.0: 5afcf98fcd03f1c2f92c3c83f58ae24323cc57fd
+actions/attest@v4.1.1: a1948c3f048ba23858d222213b7c278aabede763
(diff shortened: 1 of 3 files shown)


avifenesh:
All four addressed in 604d2827. The exhaustiveness change worked immediately — one review, all three corners of finding 1 together instead of three rounds. And finding 1 was that same class one level up: I fixed per-commit → per-branch, not per-PR.

**1, per-branch-name ≠ per-PR.** Both failure modes measured:

| check | result |
|---|---|
| remote heads surviving | **66**, incl. recurring `codex/*` names |
| `pull_requests[].number` | populated for the open PR, **empty** once merged — so it can't be the filter, as you said |
| #1290's branch, PR-scoped | 15 (correct: all 15 were that PR) |
| same branch, now-floor | **0** — the reused-name case gets a fresh budget |

Now branch + `head_repository.full_name` + a `created_at` floor at the PR's creation. `HEAD_REF` URL-encoded via `jq '@uri'`; verified `feat/a+b` → `feat%2Fa%2Bb`, since `+` would have decoded to a space and silently switched the cap off.

**2, dead guards.** Right on both counts — `awk END {print total+0}` always prints and exits 0, and `||` binds to the pipeline. Replaced with a check that can fire. Your point about the cost of fail-open growing is what made this worth fixing rather than deleting.

**3, orphaned `HEAD_SHA`.** Put back in the log line rather than removed, for the reason you gave.

**4, the missing test — best finding here.** A step named "Lint workflow files" linted one file, so every workflow change in this thread had no CI gate. One correction: `args: .github/workflows/` **fails** — `read .github/workflows/: is a directory`, verified locally with v1.7.12. Dropped `args` entirely instead; no-arg discovers every workflow. All pass, and since actionlint runs shellcheck over `run:` blocks it now covers the shell in this very step.

**Out-of-diff item taken too:** the README SHA table listed actionlint v1.7.1 and a stale claude-code-action SHA. Refreshed both against the real pins — small enough to fold in rather than defer.

**On your verification gap:** you said the sandbox refused every shell snippet and `gh api`, so findings 1 and 2 were reasoned from the API contract rather than measured, and that under the old prompt you'd have written them as flat assertions. That's exactly what the new section is for, and it's why I ran the measurements above rather than taking them on trust — they confirmed you on every point.

avifenesh:
The new lint gate failed on its first run — on my own step — which is the best possible confirmation it was needed.

`SC2086` on two unquoted `>> $GITHUB_OUTPUT` redirects in the run-count step the previous commit rewrote. Fixed in dcea629e.

Worth recording why I missed it locally: **actionlint only reports shellcheck findings when shellcheck is on `PATH`**, and it wasn't. My "ACTIONLINT CLEAN" was actionlint without its shellcheck pass — a green result that didn't mean what I read it as. Fetched shellcheck, re-ran, and now reproduce the CI invocation exactly. Also grepped the other workflows for the same pattern; none had it.

So the sequence was: you found that workflow changes had no CI gate → I widened the gate → the gate immediately caught a bug in the change that widened it. Under the old config that shell error would have merged unnoticed, since the step only ever read `docs-site.yml`.

avifenesh:
All four addressed in 5f36b1a4. First: you were right that `dcea629e` fixed only the SC2086 quotings — my reply claiming items 2-4 were done was wrong, and you checked rather than took it.

**1, duplicate skip notice (new, mine).** The sharpest finding here. Per-commit that branch was effectively unreachable; per-PR the counter is monotonic, so every later push would post another identical comment — ~10 on #1290. Marker-guarded now, once per PR, and the text points at `@claude` for another look. Same class as everything else in this thread: I widened the counter's scope and didn't look at what else that scope reached.

**2, dead guard.** Byte-identical, as you said, with a comment describing a fix the code didn't implement. `gh` now runs inside the `if` so its status is the one tested, awk sum in the `else`. Kept your correction that mid-pagination under-counts and fails open — right direction for a cost guard, and now stated as such.

**3, lint gate unreachable.** This is the one I'd have sworn was fixed. Dropping `args` made actionlint check every workflow, but `docs-site.yml`'s own `paths:` still listed only itself — so a PR touching another workflow triggered zero runs. The gate existed and could not fire. Both filters now `.github/workflows/**`, and I corrected the past-tense comment claiming it was already solved.

**4, SHA table.** Audited rather than patched: **11 stale, 6 used-but-unlisted, 3 listed-but-unused — 5 of 19 correct.** Your count was essentially right. Regenerated from the actual `uses:` pins; 22 of 22 now match, and the header states the workflows are authoritative on disagreement so the table can't silently rot again.

Verified: all workflows parse, `actionlint` + `shellcheck` clean.

Noting this was review 3 of 3 — the cap this PR introduces now binds. If more is needed I'll trigger it explicitly with `@claude`, which is exactly the escape hatch the new notice advertises.

Issue #1288, opened 2026-07-31
Audit the six A-tier spec-drift sources (Cursor, Copilot, Cline)


avifenesh:
The six A-tier sources tracked in `.github/spec-baselines.json` all show drift and have never been audited against their current upstream docs. Their baselines were set 2026-02-04.

- `cursor-rules` — https://cursor.com/docs/rules (CUR-001..009)
- `cursor-hooks` — https://cursor.com/docs/hooks (CUR-010..013)
- `cursor-subagents` — https://cursor.com/docs/subagents (CUR-014..015)
- `cursor-environment` — https://cursor.com/docs/cloud-agent/setup (CUR-016)
- `github-copilot` — https://docs.github.com/en/copilot/customizing-copilot (COP-*, CP-SK-001)
- `cline-rules` — https://docs.cline.bot/features/cline-rules/overview (CLN-*, CL-SK-*)

Their hashes were deliberately **not** re-cut in #1287 — doing so would have asserted a review that never happened. They stay flagged until audited.

## Suggested approach

Same method that worked for S-tier in #1281: one reader per source, diffing the live doc against the validator and `rules.json`, classifying each finding as MISSING_FIELD / FALSE_POSITIVE / STALE_VALUE / MISSING_RULE / NO_ACTION, then verifying every claim against the built binary before changing code.

Two things that pass worth repeating:

1. **Reproduce before fixing.** Several S-tier "findings" only became real once run through the CLI; one turned out to need fixing in two places (detector *and* validator) because the unit test passed while the CLI stayed silent.
2. **Assert both directions in tests.** Every fix there pins the documented value as accepted *and* the genuinely invalid one as still rejected.

Note the rule-ID mappings for `github-copilot` and `cline-rules` in `spec-baselines.json` were wrong until #1283 (they listed `GH-*` and `CLINE-*`, which never existed here); they now derive from each rule's own `evidence.source_urls`.

Once a source is reviewed and its findings fixed, re-cut just that hash — the workflow's `update_baselines: true` mode prints them, so they don't need hand-computing.

PR #1292, opened 2026-07-31
feat: add XP-009 for Codex's cumulative instruction-chain cap (#1289)

### LINKED ISSUE #1289: XP-007: project_doc_max_bytes cap is cumulative, not per-file
XP-007 checks each `AGENTS.md` against Codex CLI's 32768-byte limit in isolation, but the documented cap is **cumulative across the whole discovery chain**.

From the Codex AGENTS.md guide (https://learn.chatgpt.com/docs/agent-configuration/agents-md):

> Codex ... stops adding files once the **combined** size reaches the limit defined by `project_doc_max_bytes` (32 KiB by default)

Per-directory discovery order is `AGENTS.override.md`, then `AGENTS.md`, then `project_doc_fallback_filenames` — all drawing on the same budget, at most one file per directory.

## Impact

A project whose instructions are split across several mid-size files stays silent under XP-007 while Codex is actually truncating it. Example: four `AGENTS.md` files at 10 KB each in a root→cwd chain totals 40 KB, exceeding the cap, and every file passes the current per-file check.

This is a **missed detection**, not a false positive, which is why it wasn't fixed in the #1281 sweep — the byte value (`CODEX_BYTE_LIMIT = 32_768`) and the per-file check are both correct as far as they go. `AGENTS.override.md` was added to the checked filenames in #1286.

## Why it needs design work

XP-007 currently runs as a per-file validator (`crates/agnix-core/src/rules/cross_platform.rs`), and the cumulative check needs the whole root→cwd chain at once. That means moving it (or adding a companion) to project-level validation in `rules/project_level.rs`, alongside AGM-006 and XP-004/005/006, plus deciding:

- Which file does the diagnostic report on — the one that crosses the threshold, or the project root?
- Should `project_doc_fallback_filenames` from the Codex config feed the chain? agnix already parses that key under CDX-006.
- Does the existing per-file XP-007 stay as-is, or become the cumulative rule with a new ID?

Current limitation is documented in the XP-007 entry in `knowledge-base/VALIDATION-RULES.md` so it isn't silently carried.

avifenesh:
Closes #1289. Also clears #1282 (Cline v4.1.2 triage) since it needed no rule change.

## XP-009: the cap is cumulative, not per-file

From the Codex AGENTS.md guide:

> Codex … skips empty files and **stops adding files once the combined size reaches the limit** defined by `project_doc_max_bytes`

XP-007 checks each `AGENTS.md` in isolation, so a project split across several mid-size files gets truncated with **every per-file check passing**. This was the dimension deferred in #1286 and recorded in XP-007's own docs rather than left implicit.

The new project-level rule models the documented discovery order:

> "Starting at the project root (typically the Git root), Codex walks down to your current working directory" · "Codex includes at most one file per directory" · "it checks for `AGENTS.override.md`, then `AGENTS.md`, then any fallback names" · "Codex concatenates files from the root down"

So: root-down, one file per directory, override preferred, summed root-first, reported on the file where the running total crosses the limit — that file and everything deeper is what Codex drops. It reuses the paths the project walk already collected, so it adds no traversal.

**Deliberately conservative.** `project_doc_fallback_filenames` lives in the user's Codex `config.toml`, not anything agnix reads, so a project using fallbacks has a longer real chain than this models. That makes the check under-report rather than over-report — the right direction for a SHOULD-level warning, and stated in the rule docs rather than left as a surprise.

## The tests found a second bug

`AGENTS.override.md` was **not in `is_instruction_file()`**, so it was invisible to every cross-platform rule. Which means the `AGENTS.override.md` handling added to XP-007 in #1286 **could never fire** — I shipped an unreachable fix.

Caught by a test asserting a small override shadows a 40 KB `AGENTS.md`: it failed because only the `AGENTS.md` was ever collected. Fixed at the source, so XP-007's override handling now works too.

Tests assert both directions:

| case | expected |
|---|---|
| 3 × 14 KB chain (42 KB) | XP-009 fires, **XP-007 stays quiet** — that's the gap |
| 2 × 6 KB chain (12 KB) | clean |
| small override + 40 KB shadowed `AGENTS.md` | clean — shadowed bytes don't count |
| `[[overrides]]` on the reported file | suppressed |

## Cline v4.1.2 (#1282)

Reviewed **all three** releases since the `v4.0.12` baseline, not just the newest one the issue named:

- `v4.1.0` — A/B packaging: one VSIX with both extensions plus a loader
- `v4.1.1` — removes internal MCP server-key machinery from `McpHub`
- `v4.1.2` — settings-page label showing which variant is active

All three match this tool's declared-irrelevant list (`Webview UI updates`, `cline-core internal…`, `Provider plumbing`). None touches `.clinerules` parsing, workflow/hook/skill layouts, or skill frontmatter. Confirmed agnix validates `.clinerules` and skill files, not Cline's MCP internals, so v4.1.1 is out of scope too.

Baseline advanced to `v4.1.2` and the `RESEARCH-TRACKING.md` review date updated. No rule changes.

## Verification

5060+ tests / 37 binaries green · eval 61/61 · self-lint clean · `cargo fmt --check` and `cargo clippy --all-targets` clean · bookkeeping, locale sync and rule counts in sync at **443** · `actionlint` + `shellcheck` clean.

Rule count 442 → 443.


avifenesh:
Both blocking findings fixed in 9617d243. Reproduced each before touching code.

**1, sibling subtrees.** You were right, and the reproduction matched your numbers exactly:

```
root 10 KB + pkg-a 12 KB + pkg-b 12 KB
real chains: root->pkg-a = 22 KB, root->pkg-b = 22 KB   (both fit)
before:      XP-009 on pkg-b, "reaches 34837 bytes across 3 file(s)"
after:       0 diagnostics
```

The argument that settled it was yours: it contradicts AGM-006 two hundred lines above in the same file, which recommends splitting across nested directories *precisely* to stay under this cap. A rule punishing the layout its sibling recommends is wrong regardless of the arithmetic.

Also confirmed your measurement that this repo was ~5 KB from tripping its own rule — 19 files, **27,641** of 32,768. It now reports nothing.

Now one chain per leaf directory, summed independently, reported on the file where that chain crosses, deduped so a shared root isn't reported once per descendant.

**2, override pairs.** Also correct, and a consequence of my own fix: adding `AGENTS.override.md` to `is_instruction_file()` for XP-007/XP-009 fed it to XP-004/005/006, which then compared an override against the file it shadows. Codex "includes at most one file per directory" and checks the override first, so the shadowed file is never loaded. Shadowed `AGENTS.md` files are dropped from those detectors' candidate sets, using the same filter-at-input approach the `[[overrides]]` handling already uses.

Verified a genuine `CLAUDE.md`-vs-`AGENTS.md` conflict still fires, so this is an exclusion rather than a disablement.

Four regression tests cover both directions of each.

One note on process: I'd initially pushed the A-tier work (#1288) onto this branch, which conflated two issues. Split it out to #1293 so each is reviewable on its own — this PR is back to XP-009 plus the Cline baseline only.

avifenesh:
All three fixed in fedcaa30.

**1, attribution.** Reproduced your exact numbers — and confirming your framing, this was the sibling bug one step along: N reports against the wrong file instead of one report with the wrong total.

```
30 KB root + 20 x 5 KB packages
before: 20 x XP-009, one per pkgNN/AGENTS.md
after:   1 x XP-009 on the root
```

Now attributed to the largest contributor in the chain, tie-broken toward the shallowest. A deep file that dominates its chain is still named, so it isn't just "always blame the root" — there's a test for that direction too.

Your point that **no test could see it** was the useful half: `xp009_does_not_sum_across_sibling_subtrees` uses siblings that stay *under* the cap, so the crossing case was unexercised. Added both.

**2, three conventions for one filename.** Correct. `is_instruction_file()` matched case-insensitively while my shadow filter and the chain precedence matched exactly, so a lowercase `agents.override.md` was collected as an instruction file, shadowed nothing, and never entered the chain — leaving the 40 KB file it should have hidden in the sum. All three now use `eq_ignore_ascii_case`, tested against both spellings.

**3, XP-007 asymmetry.** Recorded in XP-007's rule docs rather than left implicit. You're right that two rules in one PR were taking opposite positions on the same file, and right that it isn't fixable from a per-file validator — it needs the project-level shadowing model XP-009 has.

One process note: my first attempt to reproduce (1) through the CLI showed zero diagnostics and I nearly wrote it off. The temp fixtures were being excluded by the directory walk; reproducing through `validate_project` in a test showed all three immediately. Worth flagging since a CLI-based "cannot reproduce" would have been wrong.

Verified: 37 binaries green, eval 61/61, fmt + clippy `-D warnings` clean, bookkeeping and locales in sync.

PR #1293, opened 2026-07-31
fix: A-tier spec drift for Cursor, Copilot and Cline (#1288)

### LINKED ISSUE #1288: Audit the six A-tier spec-drift sources (Cursor, Copilot, Cline)

avifenesh:
Closes #1288 — the six A-tier sources that had never been audited, and whose baselines were deliberately left stale in #1287 for exactly that reason.

**Nine false positives**, each reproduced against the doc's own example before fixing and re-verified after.

## Copilot

**COP-005 rejected the documented value and autofixed it wrong.**

> "add the `excludeAgent` keyword to the frontmatter block. Use either `"code-review"` or `"cloud-agent"`."

agnix listed `coding-agent`, which was never upstream. So the doc's own value errored — and worse, the unsafe autofix rewrote a *correct* value into the invalid one via `find_closest_value`.

**56 occurrences across 22 files.** Fixing only the const would have left every user-facing message wrong, so this covers the validator, all 12 locale files, both `rules.json` copies, and the knowledge-base docs. `coding-agent` is retained as a deprecated alias — users who followed agnix's own wrong advice shouldn't break — ordered last so the fixer prefers the documented spelling.

**COP-003 rejected comma-separated `applyTo`.**

> "You can specify multiple patterns by separating them with commas." — with `applyTo: "**/*.ts,**/*.tsx"` as the example

The whole string went to `glob::Pattern`. The notable part: a correct brace/bracket-aware splitter **already existed** with 18 passing tests, marked `#[allow(dead_code)] // reserved for future use` and never called. The fix was wiring, not new logic.

## Cursor

**CUR-004 — same comma bug**, same upstream wording ("Separate multiple patterns with commas"). Worth calling out *why it survived*: it was **intermittently** wrong. The doc's literal row `docs/**/*.md, docs/**/*.mdx` happens to parse as one pattern, while `src/**, tests/**` does not — so a fixture built from the doc's own example would have passed.

**CUR-010 required `version`** — documented as `| version | number | 1 | Config schema version |`, and several of the doc's examples omit it. Now optional, and typed *number* rather than integer, so `1.0` is valid.

**CUR-012 required `command` on prompt hooks.** The doc's prompt example has only `type`, `prompt`, `timeout` — "Prompt hooks use an LLM to evaluate a natural language condition". CUR-018 already checked `prompt` for those.

**CUR-016 required `install`.** I fetched the published schema rather than trusting the page:

```
root required: NONE          root unevaluatedProperties: False
def common: required=NONE    def container: required=[]
common props: install, name, ports, repositoryDependencies, start, terminals, user
'update' present? False
terminals items oneOf[0] type=array required=['command']
                oneOf[1] type=object required=['command']
```

So: nothing is required, the setup page's snapshot example legitimately omits `install`, terminal entries need only `command` (both `oneOf` branches), and `update` is **not in the schema** — which sets `unevaluatedProperties: false`, making it invalid. It's now reported as renamed to `install` rather than silently accepted.

**Timeout units** — two suggestions said milliseconds; the doc says "Execution timeout in **seconds**".

## Cross-tool

**CC-MEM-014 fired on Cursor `.mdc` files**, reporting them as `CLAUDE.md has N non-empty lines, exceeding the recommended 200 line limit` — wrong filename, wrong threshold (Cursor documents 500), wrong tool attribution. `ClaudeMdValidator` is registered for `FileType::CursorRule` for its generic prose checks; the Claude-specific line rule is now gated on the file actually being a Claude memory file.

**18 rules had redirecting evidence URLs** — Copilot 301s to `/how-tos/provide-context`, Cline 308s to `/customization/cline-rules`. Both verified with `curl` and refreshed in `rules.json` and `spec-baselines.json`, so those two sentinel sources can match again.

## Where the two audits disagreed

One auditor reported comma-globs as a five-tool class including Cline and Claude rules. I checked each upstream doc separately rather than applying it as a blanket change:

| tool | field | verdict |
|---|---|---|
| Cursor `globs` | scalar string | **comma-separated** — fixed |
| Copilot `applyTo` | string | **comma-separated** — fixed |
| Cline `paths` | YAML array | **not** comma-separated; scalar form already errors — no change |
| Claude rules `paths` | YAML array | array form documented — no change |

Splitting on commas for Cline or Claude would have been wrong. The docs settled it per tool.

## Deliberately not done

User-global paths (`~/.copilot/`, `~/Documents/Cline/Rules`, `~/.agents/AGENTS.md`) are outside the project tree agnix walks. Adding rules there is a scope decision, not a drift fix — worth a separate issue if wanted.

Baseline hashes stay stale until this merges; re-cutting them is a separate reviewable change, same as #1287.

## Verification

5060+ tests / 37 binaries green · eval 61/61 · self-lint clean · `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean · bookkeeping, locale sync and rule counts in sync at 442 · `actionlint` + `shellcheck` clean.

Every fix reproduced before and re-verified after; new tests assert both directions (documented value accepted **and** the genuinely invalid one still rejected).


avifenesh:
Fixed in 6750bf10. One of your two "wrong" findings was right and one wasn't — details below, with the measurement for each.

**1, the autofix could still write `coding-agent`. Correct, and it's the PR's own headline bug through a different input.** Reproduced with `--fix-unsafe`:

```
before:  excludeAgent: "coding"  ->  autofix wrote "coding-agent"
         (while the same diagnostic said: Valid values: 'code-review', 'cloud-agent')
after:   excludeAgent: "coding"  ->  reported, no fix offered
```

Split into `DOCUMENTED_AGENTS` (fix candidates) and `ACCEPTED_AGENTS` (input tolerance), exactly as you suggested. `coding` now matches no documented value so no fix is guessed — the right outcome.

Your diagnosis of *why it survived* was the useful part: all three new COP-005 tests asserted a diagnostic fired, none asserted the replacement **text**. Added a test that checks the fix text directly.

Also took the softer point — the legacy spelling now reports at **info** level instead of being accepted silently. A user who followed agnix's old advice was shipping a file Copilot won't honor and hearing nothing.

**2, terminals `oneOf` — I don't think this one holds.** I re-fetched the schema and dumped the structure:

```json
"terminals": { "type": "array", "items": { "oneOf": [
  { "type": "array", "items": { "type":"object", "required":["command"] } },
  { "type": "object", "required": ["command"] } ] } }
```

The `oneOf` is on **`items`**, not on the `terminals` value, and branch 1 *is* an array of command-bearing objects. So the branches aren't inverted and there is an array-of-arrays branch.

**But you found a real gap next to it:** I accepted the array branch without checking inside, so `[[{"name":"x"}]]` passed with no `command`. Now validated, with tests for both directions plus the scalar case that matches no branch.

If you're reading the schema differently, point me at the path — I'm working from `cursor.com/schemas/environment.schema.json` fetched today, and I'd rather be corrected than leave a wrong reading in a comment.

Verified: 37 binaries green, eval 61/61, fmt + clippy `-D warnings` clean, bookkeeping and locales in sync.

PR #1431, opened 2026-08-26
feat(pypi): publish agnix wheels to PyPI

### LINKED ISSUE #1430: Publish on PyPI
Would you consider publishing **agnix** on [PyPI](https://pypi.org/)? Thanks 🙏

avifenesh:
## What

Adds a PyPI distribution: `pip install agnix` and `uvx agnix .` join the npm, Homebrew, and cargo channels.

New `pypi/` directory:

| File | Purpose |
|---|---|
| `pyproject.toml` | Single source of truth for the wheel metadata (no PEP 517 backend - see below) |
| `agnix/__init__.py` | Python API mirroring `npm/index.js`: `run`, `lint`, `version`, `binary_path` |
| `agnix/__main__.py` | `python -m agnix`, execs the binary |
| `build_wheels.py` | Repacks release archives into one platform wheel per target |
| `test/test_build_wheels.py` | Unit tests for the builder |

## Why repack instead of maturin

`release.yml` already cross-builds `agnix` for all five targets, checksums each archive, and attests it. Building through maturin would compile the same Rust a sixth time per platform and drop the attestation chain. `build_wheels.py` verifies each archive against its `.sha256` sidecar - the same check `npm/install.js` does - and wraps the binary as-is.

## Details worth reviewing

- **manylinux tags are derived, not hardcoded.** Each glibc wheel's tag comes from the highest `GLIBC_2.x` symbol the binary actually references (currently 2.18, well under the 2.31 floor `check-glibc-floor.sh` enforces). A toolchain or cross-image bump therefore cannot silently ship a wheel claiming wider compatibility than the binary has. `manylinux_2_17` additionally gets the legacy `manylinux2014` alias.
- **Binary location.** It ships under `agnix-<version>.data/scripts/`, the only wheel location whose executable bit survives installation, and lands as `agnix` on PATH directly. There is deliberately no `console_scripts` entry named `agnix`, which would collide with it. `external_attr` carries a full `st_mode` including `S_IFREG`, because pip's `zip_item_is_executable()` checks `S_ISREG` before it looks at permission bits - permission bits alone install the binary non-executable.
- **Platform coverage.** manylinux + musllinux x86_64, manylinux aarch64, macOS arm64, Windows x86_64. Intel macOS is not covered because no such release binary exists; that platform falls back to `cargo install agnix-cli`, as documented.
- **Wheels are reproducible** (fixed zip timestamps), so rebuilding a release produces byte-identical artifacts.

## Release wiring

New `pypi` job in `release.yml`, `needs: release`, prereleases skipped, mirroring the npm job:

1. Sync versions from the tag
2. Download the release artifacts
3. Build wheels
4. **Install the Linux wheel and assert `agnix --version` equals the tag** before any upload
5. `twine check`, then `twine upload --skip-existing` (idempotent on re-run, matching the crates.io helper)

`scripts/sync-versions.sh` now also syncs `pypi/pyproject.toml` and `pypi/agnix/__init__.py`.

## Blocking on

A `PYPI_API_TOKEN` repository secret. Everything else is in place; the job will fail at the upload step until it is set.

## Verification

Built wheels from the real published v0.49.0 archives and installed one in a clean venv:

- `agnix --version` -> `agnix 0.49.0`
- `python -m agnix --version` -> `agnix 0.49.0`
- `agnix.version()`, `agnix.lint(...)`, `agnix.binary_path()` all work
- `uvx --from <wheel> agnix --version` works
- `twine check` PASSED on both wheels
- 16 builder tests pass; `agnix .` on this tree reports no issues

## Docs

README install block, `website/docs/installation.md` (new pip/uv section), `docs/RELEASING.md` (PyPI row and publish step), CHANGELOG - all in this commit.

Closes #1430.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,12 +78,14 @@ jobs:
 
       - name: Release install path tests
         if: matrix.os == 'ubuntu-latest'
-        # Covers the glibc floor gate the release build runs and the npm
-        # installer's static-build fallback. Neither ships inside the Rust
-        # workspace, so nothing else exercises them (see #1371).
+        # Covers the glibc floor gate the release build runs, the npm
+        # installer's static-build fallback, and the PyPI wheel builder. None of
+        # them ship inside the Rust workspace, so nothing else exercises them
+        # (see #1371).
         run: |
           bash scripts/check-glibc-floor.test.sh
           node npm/test/install-helpers.test.js
+          python3 -m unittest discover -s pypi/test
 
       - name: Schema sync check
         if: matrix.os == 'ubuntu-latest'
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -445,6 +445,73 @@ jobs:
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
 
+  pypi:
+    name: Publish PyPI Wheels
+    needs: release
+    runs-on: ubuntu-latest
+    # Skip prereleases
+    if: ${{ !contains(github.ref, '-') }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - uses: actions/setup-python@5fda3b95a4ea91299a34e894583c3862153e4b97 # v7.0.0
+        with:
+          python-version: '3.x'
+
+      - name: Sync all versions from tag
+        run: |
+          set -euo pipefail
+          TAG_VERSION="${GITHUB_REF#refs/tags/v}"
+          bash scripts/sync-versions.sh "${TAG_VERSION}"
+
+      # The wheels repack the archives this release already built, verified
+      # against their .sha256 sidecars, instead of compiling the same Rust a
+      # sixth time.
+      - name: Download release artifacts
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          path: artifacts
+          merge-multiple: true
+
+      - name: Build wheels
+        run: |
+          set -euo pipefail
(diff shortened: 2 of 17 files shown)


avifenesh:
Both findings addressed in 4c15d51a.

**1. `--target ClaudeCode` was rejected, and `lint()` hid it.** Confirmed against `TargetArg` in `crates/agnix-cli/src/main.rs` - accepted values are `generic`, `claude-code`, `cursor`, `codex`, `kiro`. clap exits 2 with empty stdout, so the old `json.loads` fallback returned `{"files": [], "summary": {"errors": 0, ...}}`, i.e. a clean lint.

- `lint()` now raises `AgnixError` (new, `RuntimeError` subclass) carrying the exit code and stderr when a JSON report was requested but not produced.
- A non-`json` `--format` returns `{"raw": ...}` instead of being run through the JSON parser at all, so the raise means "something went wrong", not "you asked for text".
- Docs corrected in `pypi/README.md` and `website/docs/installation.md`.

Verified against the real v0.49.0 binary:

```
lint json: {'errors': 0, 'warnings': 0, 'info': 0}
rejected arg raises AgnixError: agnix exited 2 without JSON output: error: invalid value 'ClaudeCode' for '--target <TARGET>'
  [possible values: generic, claude-...
```

**2. Stale rule count.** `pypi/README.md` now says 451, and so does `npm/README.md`, which is where I copied 405 from - the npm long description had drifted unguarded for several releases. Both wrapper READMEs are now in `sync-rule-bookkeeping.js`'s `COUNT_FILES`, so the next rule addition fails CI instead of shipping a stale number to two registries. `node scripts/sync-rule-bookkeeping.js --check --skip-docs` passes.

Also added `pypi/test/test_api.py`: argument construction (including that `--target` is omitted when no tool is passed), the raise on non-JSON output, raw passthrough, and binary resolution across the scripts dir / user scheme / PATH fallback. 25 tests total in `pypi/test`, all passing.

PR #1441, opened 2026-08-27
feat(pre-commit): let pre-commit install agnix itself

### LINKED ISSUE #1439: Use `language: python` for `pre-commit` hooks
Could you change the hooks in `.pre-commit-hooks.yaml` from `language: system` to `language: python`? This would let `pre-commit`/`prek` automatically install and manage `agnix`, instead of requiring a manual system installation.

avifenesh:
Closes #1439.

The three hooks move from `language: system` to `language: python`, so `pre-commit` and `prek` create a virtualenv and install the `agnix` wheel matching the pinned `rev` - instead of failing unless the user already ran `cargo install agnix-cli`. Publishing to PyPI (#1431) is what made this possible, and @pygarap filed this the same day.

## Why a Rust repo now has a root `pyproject.toml`

A `language: python` hook makes pre-commit clone the repo and `pip install .` from its **root**, so the clone has to be pip-installable. The root `pyproject.toml` is a shim with no Python code of its own:

```toml
[project]
name = "agnix-pre-commit"
version = "0.51.0"
dependencies = ["agnix==0.51.0"]
```

Named differently from the published `agnix` so the two can never collide, and never published itself. The wheel puts the binary in the venv's scripts directory, which is exactly where `entry: agnix` resolves. This mirrors how other Rust tools do it (ruff's mirror repo), just in-repo rather than as a separate mirror.

`scripts/sync-versions.sh` moves the shim's version **and** its `agnix==` pin together, so `rev: vX.Y.Z` always runs agnix X.Y.Z rather than drifting to latest. Verified both lines move by syncing to a throwaway version and back.

## Verified end to end, not just by inspection

Scratch repo, `pre-commit 4.6.2`, hook repo pinned at this commit:

```
agnix ....................... Failed   # 2 warnings under --strict, correctly non-zero
agnix (with auto-fix) ....... Passed
agnix-schema ................ Passed
```

This machine already has `agnix` on PATH at `~/.cargo/bin/agnix`, so "it ran" alone would prove nothing. Checking the hook environment:

```
$ .../py_env-python3.12/bin/pip list | grep agnix
agnix            0.51.0
agnix-pre-commit 0.51.0

$ readlink -f .../py_env-python3.12/bin/agnix
/home/.../py_env-python3.12/bin/agnix        # a real 9.6MB binary, not the cargo one
```

The venv holds its own binary from the PyPI wheel; PATH was not consulted.

## Platform caveat, documented

Wheels cover Linux x86_64 (glibc + musl), Linux aarch64, macOS Apple silicon, and Windows x86_64. Intel macOS has no release binary, so `docs/CONFIGURATION.md` now documents pinning `language: system` in your own config as the fallback - and drops the stale `rev: v0.37.5` examples while it is there.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -2,7 +2,7 @@
   name: agnix
   description: Lint AI agent configuration files
   entry: agnix
-  language: system
+  language: python
   types: [file]
   files: |
     (?x)^(
@@ -25,7 +25,7 @@
   name: agnix (with auto-fix)
   description: Lint and auto-fix AI agent configuration files
   entry: agnix
-  language: system
+  language: python
   types: [file]
   files: |
     (?x)^(
@@ -53,7 +53,7 @@
     fails the commit if the working tree is dirty afterward. Override the
     target path with `args: [--output=<path>, --fix]`.
   entry: agnix schema
-  language: system
+  language: python
   pass_filenames: false
   always_run: true
   args: [--fix]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- **pre-commit hooks install agnix themselves**
+  ([#1439](https://github.com/agent-sh/agnix/issues/1439)). The three hooks in
+  `.pre-commit-hooks.yaml` move from `language: system` to `language: python`,
+  so `pre-commit` and `prek` create a virtualenv and pull the `agnix` wheel that
+  matches the pinned `rev` instead of requiring a manual `cargo install` on
+  PATH. This is what the new PyPI distribution makes possible. A root
+  `pyproject.toml` exists solely for that install: pre-commit clones the repo
+  and `pip install .`s it, so the clone has to be pip-installable, and this shim
+  declares the matching `agnix==<version>` as its only dependency.
+  `scripts/sync-versions.sh` moves the shim's version and its pin together, so a
+  hook pinned at `rev: vX.Y.Z` always runs agnix X.Y.Z. Platforms without a
+  wheel can still pin `language: system` in their own config, which the
+  configuration guide now documents.
+
 ### Added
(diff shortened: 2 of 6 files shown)


avifenesh:
Fixed in 836b8468. You are right that the snippet contradicted the sentence under it: `rev: v0.51.0` is the released tag, and it still ships `language: system` hooks, so a copy-paste of that example produces exactly the PATH failure #1439 asked to remove.

All three examples now pin `v0.52.0` - the first release that will carry these hooks - and the docs say so explicitly:

> This needs `rev: v0.52.0` or later. Earlier tags ship `language: system` hooks, which require an `agnix` already on your PATH.

v0.52.0 is the release going out right after this merges, so the tag lands before anyone can follow the docs to it.

Also confirming your version-triangle and release-isolation checks match what I verified locally: `build_wheels.py` reads `PYPI_DIR / pyproject.toml` rather than CWD, and the twine upload globs `pypi/dist/*.whl`, so the root shim cannot leak into the published package.

avifenesh:
Added the guard in 958c7ca2. You are right that this was the one version-sync invariant with nothing enforcing it, and the sibling guards you listed are exactly the precedent - `sync-rule-bookkeeping.js`'s own comment records the 4-minor-version drift that motivated its.

Extended `pypi/test/test_build_wheels.py`, which already runs in CI at ci.yml:88, rather than adding a new gate:

```python
def test_precommit_shim_pins_the_workspace_version(self):
    expected = workspace_version()
    shim = bw.read_metadata(bw.REPO_ROOT / "pyproject.toml")
    self.assertEqual(shim["version"], expected)
    self.assertEqual(shim["dependencies"], [f"agnix=={expected}"])
    self.assertNotEqual(shim["name"], "agnix")   # must not collide with the published package
```

The workspace-version lookup moved to a module-level helper so both version tests share it.

Verified the guard actually bites rather than trusting it green - pinned `agnix==0.50.0` against the 0.51.0 workspace:

```
drifted pin -> FAILS as intended
AssertionError: Lists differ: ['agnix==0.50.0'] != ['agnix==0.51.0']
: the pre-commit shim must pin exactly the agnix version it ships with
restored -> passes
```

26 tests in `pypi/test`, all passing.

PR #1446, opened 2026-08-27
fix(pre-commit): match every path agnix itself validates

### LINKED ISSUE #1444: Add Agent Plugins v1.1.0 Portable Paths to Pre-Commit Hooks
Update `.pre-commit-hooks.yaml` so both `agnix` and `agnix-fix` support Agent Skills in any directory and Agent Plugins v1.1.0 files:

```diff
-      SKILL\.md|
+      (?:.*/)?SKILL\.md|
+      (?:.*/)?plugin\.json|
+      (?:.*/)?mcp\.json|
```

Using `(?:.*/)?SKILL\.md` automatically covers all official and future Agent Skills locations without enumerating client-specific directories.

Reference: https://github.com/agentplugins/agent-plugins-spec/blob/main/spec/1.1.0.md

### LINKED ISSUE #1445: Support All GitHub Copilot Repository Custom Instruction Paths in Pre-Commit Hooks
Update `.pre-commit-hooks.yaml` so both `agnix` and `agnix-fix` include all repository custom instruction paths officially supported by GitHub Copilot:

- `.github/copilot-instructions.md`
- `.github/instructions/**/*.instructions.md`
- `AGENTS.md` at any directory depth
- Root `CLAUDE.md`
- Root `GEMINI.md`

In particular, generalize the existing `AGENTS.md` match:

```diff
-      AGENTS\.md|
+      (?:.*/)?AGENTS\.md|
```

and add support for path-specific instructions and `GEMINI.md`.

Reference: https://docs.github.com/en/copilot/how-tos/copilot-on-github/customize-copilot/add-custom-instructions/add-repository-instructions

avifenesh:
Closes #1444, closes #1445.

Both issues are the same underlying gap. The hooks' `files:` pattern was `^(...)$`-anchored with root-level entries, so `SKILL\.md` matched only a repository-root SKILL.md. A skill at its normal `.claude/skills/<name>/SKILL.md` location, a nested `AGENTS.md`, a plugin manifest, a client-path `mcp.json`, or a scoped Copilot instruction file **never reached agnix - the hook reported Passed while validating nothing.**

## The fix mirrors agnix's own detector

`file_types/detection.rs` recognizes these filenames at any depth, so the hook now does too via a shared `(?:.*/)?` prefix:

| Entry | Status |
|---|---|
| `SKILL.md`, `CLAUDE[.local].md`, `AGENTS[.local/.override].md` | now any depth |
| `GEMINI.md`, `GEMINI.local.md` | **new** (#1445) |
| `plugin.json` | **new** (#1444) |
| `mcp.json` | **new** - previously only `*.mcp.json` matched (#1444) |
| `.github/instructions/**/*.instructions.md` | **new** (#1445) |
| `.github/copilot-instructions.md`, `.claude/settings*`, `.claude/agents/*.md`, `.cursor/rules/*.mdc` | kept; depth-crossing `.*` tightened to `[^/]+`/`.+` |

Matching `plugin.json` anywhere is deliberate: agnix's own directory walk treats any `plugin.json` as a plugin manifest, so the hook now agrees with `agnix .` rather than being quieter than it.

## Verified two ways

**Pattern semantics** - evaluated exactly as pre-commit does (Python `re.search` on the relative path): 29 representative paths match, 10 near-misses stay excluded (`README.md`, bare `settings.json`, `SKILL.md.bak`, `.cursor/rules/style.md`, `.github/instructions/notes.md`, ...).

**End to end** - pre-commit 4.6.2 against a repo holding a nested skill with a bad name, a package-level `AGENTS.md`, `GEMINI.md`, and a scoped instruction file missing frontmatter. All four were invisible to the old pattern; the hook now fails with their real diagnostics:

```
.github/instructions/api.instructions.md:1:0 error: Scoped Copilot instruction file missing required frontmatter
.claude/skills/review/SKILL.md:2:1 error: Name 'Review-Code' must be ...
GEMINI.md:1:0 warning: Missing project context section
packages/web/AGENTS.md:1:0 warning: Missing project context section
Found 2 errors, 2 warnings
```

Docs examples move to `rev: v0.52.1`, the first tag that will carry this pattern; the version note now distinguishes v0.52.0 (self-installing hooks) from v0.52.1 (full file coverage).

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -17,12 +17,47 @@
       plugin\.json|
       mcp\.json|
       [^/]+\.mcp\.json|
-      \.claude/settings\.json|
-      \.claude/settings\.local\.json|
+      mcp-[^/]+\.json|
+      opencode\.jsonc?|
+      gemini-extension\.json|
+      \.geminiignore|
+      \.cursorrules(\.md)?|
+      \.clinerules(/.+)?|
+      \.roorules|
+      \.roomodes|
+      \.rooignore|
+      \.windsurfrules|
+      \.claude/settings(\.local)?\.json|
+      \.claude/managed-settings\.json|
       \.claude/agents/.+\.md|
+      \.claude/rules/.+\.md|
+      \.claude/output-styles/[^/]+\.md|
       \.github/copilot-instructions\.md|
       \.github/instructions/.+\.instructions\.md|
-      \.cursor/rules/.+\.mdc?
+      \.github/agents/[^/]+\.agent\.md|
+      \.github/prompts/[^/]+\.prompt\.md|
+      \.github/hooks/hooks\.json|
+      \.github/workflows/copilot-setup-steps\.ya?ml|
+      \.cursor/rules/.+\.mdc?|
+      \.cursor/agents/.+\.md|
+      \.cursor/(.+/)?hooks\.json|
+      \.cursor/environment\.json|
+      \.cursor/[^/]+\.mdc?|
+      \.codex/config\.(toml|json|ya?ml)|
+      \.codex/requirements\.toml|
+      \.gemini/settings(\.local)?\.json|
+      \.gemini/agents/[^/]+\.md|
+      \.amp/settings(\.local)?\.json|
+      \.agents/(checks|skills)/.+\.md|
+      \.roo/rules[^/]*/.+\.md|
+      \.windsurf/(rules|workflows)/.+\.md|
+      \.windsurf/[^/]+\.md|
+      \.kiro/steering/.+\.md|
+      \.kiro/[^/]+\.md|
+      \.kiro/agents/.+|
+      \.kiro/hooks/[^/]+\.kiro\.hook|
+      \.kiro/settings/[^/]+\.json|
+      \.kiro/powers/.+/POWER\.md
     )$
   args: [--strict]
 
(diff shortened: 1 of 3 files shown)


avifenesh:
Both fixed in 1ebdb4f5.

**Nested subagents** - you are right, and your diagnosis of *why* is exact: the shared `(?:.*/)?` prefix absorbs directories before the literal `.claude/agents/`, so unlike the `mcp.json` entry there is nothing left to absorb the inner directories. `[^/]+` was a regression against both the old pattern and `detect_file_type`'s parent-or-grandparent `agents` arm. Restored to `.+` in both hook blocks.

**Cursor `.md` rules** - taking this as more than a parity note: CUR-020 (Ignored Plain Markdown Cursor Rule) exists precisely to catch `.md` files in `.cursor/rules` that Cursor silently ignores. An `.mdc`-only hook can never fire the rule that detects the mistake, which is the same silent-skip class this whole PR fixes. Now `\.cursor/rules/.+\.mdc?` passes both extensions.

Regex suite re-run with both classes added: `.claude/agents/frontend/react.md` and `a/b/x.md`, `.cursor/rules/style.md` at two depths, plus `.mdcx` and a bare `agents/loose.md` as exclusions - all behave. CHANGELOG notes both.

avifenesh:
Extended to full tool-scoped parity in 33211f47, taking your first option rather than scoping the claims down - with one carve-out you yourself endorsed.

**Measured against detection.rs's own tests** (150 concrete asserted paths): misses went **112 -> 39**, and every lowercase tool-dot-dir path now matches. Added beyond your list: `hooks.json` at any depth under `.cursor` (your `project/.cursor/subdir/hooks.json` case), and the direct-child `.md` files detection classifies in `.cursor`/`.kiro`/`.windsurf` (`.cursor/README.md` -> CursorRule per its tests).

**The 39 that remain are three deliberate classes, now named in the CHANGELOG instead of over-claimed:**
1. Loose fallback arms outside a tool directory - bare `settings.json`/`hooks.json`/`environment.json`/`POWER.md`/`requirements.toml` anywhere, bare `agents/`/`steering/`/`rules-*/` dirs. Matching those would drag unrelated files in; same reasoning as your own bare-`settings.json` note.
2. Uppercase directory spellings (`.GITHUB/`, `.KIRO/`). A global `(?i)` would overmatch the case-sensitive filename arms (`skill.md` != `SKILL.md` in detection).
3. `GenericMarkdown`.

The changelog sentence now reads "every tool-scoped path shape ... verified against the 150 concrete paths in detection.rs's own tests" with the exclusions listed - no more "mirrors detection.rs" / "full file coverage" over-claim, and the docs line inherits the same scoping.

15 negative controls stay excluded (`docs/POWER.md`, `.mdcx`, `opencode.json5`, `.clinerulesx`, ...). E2E re-run: `.cursorrules`, an invalid `.codex/config.toml`, a kiro steering file, and a plain-md Cursor rule - all invisible before - now fail the hook with real diagnostics, CUR-020 included.

avifenesh:
All four inline findings fixed in 94f85e85; the two wording threads were already addressed by 33211f47's rewrite (the claim is now scoped to tool-surfaces with the exclusion classes named, in both CHANGELOG and docs).

1. **Plugin `hooks/hooks.json`** - now an anywhere-entry, subsuming the `.github`-specific one; `\.cursor/(.+/)?hooks\.json` keeps Cursor's layout. `my-plugin/hooks/hooks.json` matches.
2. **Codex inversion** - you were right and the cause is instructive: my parity extraction read `detect_codex_requirements`'s `assert_ne` as a positive assertion. Now `[Cc]odex/requirements\.toml` (undotted, case-insensitive parent per detection.rs:587), and `.codex/requirements.toml` is a negative control.
3. **Bare `agents/` dirs** - `agents/[^/]+(/[^/]+)?\.md` mirrors the parent-or-grandparent contract exactly: `agents/README.md` and `agents/sub/file.md` match, `agents/a/b/deep.md` stays excluded like detection's fall-through.
4. **`.kiro/settings.json`** - added beside the directory entry.

Parity: misses 39 -> **32** of 149 test paths, every remaining one in the three documented classes (loose fallback arms outside a tool dir, uppercase spellings, generic markdown). 14 negative controls hold.

avifenesh:
The five threads from the overlapping review run target 33211f47 and were already addressed in 94f85e85, which the subsequent APPROVED review evaluated: `.kiro/settings.json` has its entry (line beside the settings/ directory one), `[Cc]odex/requirements\.toml` replaced the inverted dotted entry, `hooks/hooks\.json` covers plugin hooks at any root, `agents/[^/]+(/[^/]+)?\.md` covers bare agents directories, and the CHANGELOG names that class as covered rather than excluded.

The one live remnant was docs/CONFIGURATION.md's 'full file coverage' phrase - softened in 8d5cb4d7 to 'every tool-scoped config path (see the hook file list)', matching the changelog's scoping.

Issue #57, opened 2026-02-01
Implement cross-platform rules (XP-001..003)


avifenesh:
Implement XP-001..003 per knowledge-base/VALIDATION-RULES.md.\n\nDoD:\n- Cross-platform validator (or integrate into existing validators)\n- XP rule IDs in diagnostics\n- Fixtures (valid/invalid)\n- Tests for each rule

avifenesh:
Duplicate of #41 (XP-001..003).

Issue #55, opened 2026-02-01
Implement MCP rules (MCP-001..006) with validator and file detection


avifenesh:
Implement MCP-001..006 per knowledge-base/VALIDATION-RULES.md, including file detection for MCP configs.\n\nDoD:\n- MCP validator module + dispatch\n- MCP rule IDs in diagnostics\n- Fixtures (valid/invalid)\n- Tests for each rule

avifenesh:
Duplicate of #39 (MCP-001..006).

PR #1449, opened 2026-08-27
chore: wire the Homebrew tap into releases and drop the stale version pin

avifenesh:
Distribution loose ends found while finalizing the v0.51.0-v0.52.1 run.

## Homebrew tap was 34 releases behind

`brew install agnix` built **v0.18.0** from source. The tap's `update-formula.yml` has waited since April for a `release-published` repository dispatch this repo never sent - the tap even ships the reference workflow with a comment saying "Install this workflow in agent-sh/agnix".

- **Already fixed live:** dispatched the tap manually; the formula now builds v0.52.1 (verified sha256 committed by the tap's own workflow run).
- **This PR:** installs `.github/workflows/notify-tap.yml` so every future `release: published` event dispatches the tap automatically.
- **One owner step remains:** the workflow needs a repo-scope `TAP_DISPATCH_TOKEN` secret. Until it exists the workflow is inert (fails quietly on the missing token) and the manual dispatch remains the fallback, documented in the CHANGELOG entry.

## Vestigial in-repo formula synced

`Formula/agnix.rb` here still described v0.18.0. brew resolves the *tap* repo, not this copy, but a stale duplicate invites edits in the wrong place. Synced byte-for-byte with the tap's current v0.52.1 formula.

## CLAUDE.md/AGENTS.md version pin removed

"Current State" claimed **v0.37.3** - fifteen releases stale. `sync-rule-bookkeeping.js` updates the rule-count lines right next to it but nothing owns the version number, so it drifted through every release since. Replaced with a pointer to GitHub releases so the line cannot drift again. Both files remain byte-identical (parity-tested), and `sync-rule-bookkeeping --check` passes.

Also for the record from the same finalization sweep, outside this repo: the Zed extension was stuck at v0.48.0 upstream - fork PR #57 merged, superseded #55/#56 closed, and zed-industries/extensions#7369 opened with the v0.52.1 bump.


avifenesh:
Blocker confirmed and fixed in e2659c43 - and your diagnosis doubles as the root-cause of the original drift: with `release: published` suppressed for `GITHUB_TOKEN`-created events, notify-tap.yml would have been exactly as dead as the dispatch the tap has waited for since April.

Replaced with a `homebrew` job inside release.yml (`needs: release`, prereleases skipped, sitting with the other publishers). Dispatching from inside the run has no trigger suppression. And it reuses `COMMITTER_TOKEN` - already operating cross-repo in the zed and version-docs jobs - so `TAP_DISPATCH_TOKEN` is no longer needed and **no owner step remains**.

Proven live, not just reasoned: sent the exact dispatch the job will send (`event_type=release-published`, `client_payload[tag]=v0.52.1`) and the tap's `update-formula.yml` completed a `repository_dispatch` run on it (idempotent - formula already at v0.52.1). CHANGELOG rewritten to describe the in-release job and drop the TAP_DISPATCH_TOKEN paragraph.

PR #1450, opened 2026-08-27
fix(action): make the documented @v0 ref exist and track releases

avifenesh:
Found in the same finalization sweep as #1449.

**`uses: agent-sh/agnix@v0` has never resolved.** The README's GitHub Action snippet, four examples in `docs/CONFIGURATION.md`, and the marketplace listing all instruct `@v0` - but no `v0` tag or branch exists in the repo's history. Every user who copied the documented snippet got:

```
Error: Unable to resolve action agent-sh/agnix@v0, unable to find version v0
```

The repo's own CI never caught it because the `agnix` job exercises the action as `uses: ./`.

## Fix, in two parts

1. **Already live:** created the floating `v0` tag at the v0.52.1 release commit and verified it resolves with a fresh `git fetch origin v0` from an empty clone. The documented snippet works as of now.
2. **This PR:** the `release` job force-moves the floating major tag (`v${major}`) to each release commit before publishing, so the ref tracks the latest release the way major-version action tags are expected to. Shell derivation tested for both `v0.52.1 -> v0` and a future `v1.2.3 -> v1`. The job already has `contents: write` and a full checkout - no new permissions, no new secrets.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -249,7 +249,10 @@ jobs:
           set -euo pipefail
           VERSION="${GITHUB_REF_NAME#v}"
           MAJOR="v${VERSION%%.*}"
-          git tag -f "$MAJOR" "$GITHUB_SHA"
+          # Peel to the commit: on an annotated tag push GITHUB_SHA is the tag
+          # object, and tagging it directly would nest v0 -> tag -> commit.
+          COMMIT="$(git rev-list -n 1 "$GITHUB_REF_NAME")"
+          git tag -f "$MAJOR" "$COMMIT"
           git push -f origin "refs/tags/$MAJOR"
 
       - name: Publish release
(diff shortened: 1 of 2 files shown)


avifenesh:
Fixed in bf62ebc5: the step now carries the same `if: ${{ !contains(github.ref, '-') }}` guard as the eight publisher jobs, so a hyphenated prerelease tag builds and publishes its GitHub release as before but never moves `@v0`. Good catch - the failure mode (action users silently running beta code) is worse than the bug this PR fixes.

avifenesh:
Fixed in a71cf0fd - peeled with `git rev-list -n 1 "$GITHUB_REF_NAME"`, the same idiom the zed job uses for the same reason. Verified the failure mode locally: this repo's release tags are annotated (`git cat-file -t v0.52.1` -> `tag`), so `GITHUB_SHA` on the push event is the tag object and tagging it directly would have nested `v0 -> tag(v0.53.0) -> commit`. Note the manual `v0` I pushed earlier already points at the peeled commit (`d18c0a88`, created via `v0.52.1^{commit}`), so the live tag is correct as-is.

